PR #6, opened 2015-04-08
Adding handling for SecretData and OpaqueData objects

callidus:
Secret Data (2.2.7) can now be registered
Opaque Object (2.2.8) can now be registered

This required the adding of a NONE type to the OpaqueDataType enum
with the value of 0. This is not indicated by the standard but we
needed some value to satisfy response decoding.

Also fixed get demo

This has been tested vs a HP Atalla ESKM HSM

### DIFF
--- a/kmip/core/enums.py
+++ b/kmip/core/enums.py
@@ -366,10 +366,11 @@ class SecretDataType(Enum):
     PASSWORD = 0x00000001
     SEED     = 0x00000002
 
-
 # 9.1.3.2.10
 class OpaqueDataType(Enum):
-    pass
+    NONE = 0x80000000 # Not defined by the standard, but we need something.
+                      # The standard does say that values starting 0x8xxxxxx
+                      # are considered extensions
 
 
 # 9.1.3.2.11
--- a/kmip/core/factories/secrets.py
+++ b/kmip/core/factories/secrets.py
@@ -28,8 +28,10 @@ from kmip.core.objects import KeyMaterial
 from kmip.core.objects import KeyWrappingData
 from kmip.core.objects import KeyValue
 
+from kmip.core.secrets import OpaqueObject
 from kmip.core.secrets import PrivateKey
 from kmip.core.secrets import PublicKey
+from kmip.core.secrets import SecretData
 from kmip.core.secrets import SymmetricKey
 from kmip.core.secrets import Template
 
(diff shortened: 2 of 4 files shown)


PeterHamilton:
Thanks for the patch! I'm reviewing it now.

kmip/core/enums.py
```diff
@@ -330,10 +330,9 @@ class SecretDataType(Enum):
     PASSWORD = 0x00000001
     SEED     = 0x00000002
 
-
 # 9.1.3.2.10
 class OpaqueDataType(Enum):
-    pass
+    NONE = 0x00000000 # Not defined by the standard, but we need something
```
rellerreller: I'm afraid if we define this to 0x00 then a later standard may override this value. Have you considered using 0x8XXXXXXX? This is the pattern for extensions.

I find it bizarre in the KMIP spec that the opaque data type, which is an object that may not be interpreted by the server, requires an opaque data type. That's my two cents on the spec.

Does the kmip test documentation have an example for opaque data?

kmip/core/factories/secrets.py
```diff
@@ -132,7 +103,52 @@ def _create_template(self, value):
             return Template(value)
 
     def _create_secret_data(self, value):
-        raise NotImplementedError()
+        if value:
```
rellerreller: Should this be if value is None?

kmip/core/enums.py
PeterHamilton: Good point, something like `NONE = 0x80000000` would probably be better.
I haven't found an `OpaqueData` example in the testing documentation. Since they're non-standard, that doesn't completely surprise me but it is frustrating.

kmip/core/factories/secrets.py
PeterHamilton: See my comments on the latest commit. Forgot we needed to swap the `if` blocks around.

kmip/core/enums.py
callidus: Good point, the spec says values not defined are reserved (9.1.3.2), so lets go with  0x80000000 here

PeterHamilton:
I haven't run this against an appliance yet, I keep getting distracted with other things. I should have time this afternoon. Just wanted to give an update; I haven't forgotten.

callidus:
No problem Peter, thanks for the update

PeterHamilton:
Ok, I ran some tests against an appliance and everything worked great. :+1:

I pushed an update to master this morning for a different patch. Please rebase on master and squash all of your commits for this patch into a single commit. I'll run it through tests one last time and then merge it in.

callidus:
excellent, I'll do the rebase and squash. Thanks Peter

callidus:
OK, it seems i completely killed git trying to rebase this. Its far too easy to do and almost impossible to fix it seems, so I'm closing this and make a new pull request with the single squashed commit. Sorry for the noise

PR #7, opened 2015-04-14
Adding handling for SecretData and OpaqueData objects

callidus:
Secret Data (2.2.7) can now be registered
Opaque Object (2.2.8) can now be registered

This required the adding of a NONE type to the OpaqueDataType enum
with the value of 0. This is not indicated by the standard but we
needed some value to satisfy response decoding.

Also fixed get demo

This has been tested vs a HP Atalla ESKM HSM


rellerreller:
:+1: Looks good to me.

PR #8, opened 2015-04-20
Updating the SecretFactory to create Public and PrivateKeys with data

PeterHamilton:
This change updates the SecretFactory, adding support for the creation of PublicKeys and PrivateKeys with secret data. The intent long-term is to move away from using the SecretFactory in this way to create specific secrets and to use it instead as a simple mapping between ObjectType enumerations and default /empty secret objects.

### DIFF
--- a/kmip/core/factories/secrets.py
+++ b/kmip/core/factories/secrets.py
@@ -48,14 +48,33 @@ class SecretFactory(object):
                                                      '{2}')
 
     def create(self, secret_type, value=None):
+        """
+        Create a secret object of the specified type with the given value.
+
+        Args:
+            secret_type (ObjectType): An ObjectType enumeration specifying the
+                type of secret to create.
+            value (dict): A dictionary containing secret data. Optional,
+                defaults to None.
+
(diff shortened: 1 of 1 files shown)


kmip/core/factories/secrets.py
```diff
@@ -48,14 +48,33 @@ def __init__(self):
...
+
+        Returns:
+            secret: The newly constructed secret object.
+
+        Raises:
+            TypeError: If the provided secret type is unrecognized.
+
+        Example:
+            >>> factory.create(ObjectType.SYMMETRIC_KEY)
+            SymmetricKey(...)
```
rellerreller: I don't quite understand the example. Could you explain?

PeterHamilton: It just shows basic usage, where passing in the object type yields an empty object of that type. It could be more detailed but that would make the docstring really long. If it's confusing I can take it out.

rellerreller:
:+1:

I am ok with this merge request. We have tested it and it worked.

PR #10, opened 2015-04-24
Fixing invalid tag for the KeyMaterialStruct

PeterHamilton:


joel-coffman:
Is there a corresponding unit test for this change?

PeterHamilton:
No, but that's because there's no test suite in place yet for the `KeyMaterialStruct`, which is where we'd test valid read/write functionality. This tag bug would've shown up there pretty clearly. I can spend some time adding that test suite in.

joel-coffman:
I think a basic unit test should be provided to prevent future regressions.

PeterHamilton:
Ah, now I remember why the `KeyMaterialStruct` is there. It took me a few minutes to sift through the last minute changes I made before `0.3.0`. That `TODO` above the class saved me.

The `KeyMaterialStruct` is just a placeholder. It's slated for deletion. I didn't have time to actually implement all of the different `Transparent*Key` structures that would actually be used here, along with the requisite docs and test suites, so I put in something that would just read bytes so we'd correctly handle the response.

The correct course here would be to just fix the tag, since this class is referenced but currently unused in Barbican, and then implement all of the actual structures needed to support different `KeyFormatTypes`. Once they're in place, we can swap over Barbican and remove the `KeyMaterialStruct` for good.

I actually started this work a while back but had to shelve it because of deadlines. See `kmip/core/keys.py` for the different `KeyMaterial` objects that we should be using for each key type, and `kmip/core/factories/keys.py` for the `KeyFactory` that should be used to build them. When reading the `KeyBlock` for a response, we'd first read in the `KeyFormatType` (an enumeration). That would be passed to the `KeyFactory` to build the right `KeyMaterial` structure needed to hold the key data. That's the correct way to handle reading keys.

joel-coffman:
I still favor a simple unit test to prevent any regressions beofre `KeyMaterialStruct` is deleted.

PeterHamilton:
Not a problem. I can whip something together for it.

PeterHamilton:
Ok, basic test suite added in for the bug.

PeterHamilton:
Hmmm, looks like the pull request doesn't keep track of old versions of the branch. That....kind of sucks.

rellerreller:
I'm ok with the change.

PR #11, opened 2015-04-29
Adding support for the Certificate managed object

PeterHamilton:
This change polishes and reorganizes the implementation of the Certificate managed object and its required subclasses. It adds in documentation and test suites for all modified classes and updates the SecretFactory to support creating default Certificate objects.

### DIFF
--- a/kmip/core/factories/secrets.py
+++ b/kmip/core/factories/secrets.py
@@ -28,6 +28,7 @@ from kmip.core.objects import KeyMaterial
 from kmip.core.objects import KeyWrappingData
 from kmip.core.objects import KeyValue
 
+from kmip.core.secrets import Certificate
 from kmip.core.secrets import OpaqueObject
 from kmip.core.secrets import PrivateKey
 from kmip.core.secrets import PublicKey
@@ -68,7 +69,7 @@ class SecretFactory(object):
             SymmetricKey(...)
         """
         if secret_type is ObjectType.CERTIFICATE:
-            return self._create_certificate(value)
+            return self._create_certificate()
         elif secret_type is ObjectType.SYMMETRIC_KEY:
             return self._create_symmetric_key(value)
         elif secret_type is ObjectType.PUBLIC_KEY:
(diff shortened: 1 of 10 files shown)


kmip/core/factories/secrets.py
```diff
@@ -68,7 +69,7 @@ def create(self, secret_type, value=None):
             SymmetricKey(...)
         """
         if secret_type is ObjectType.CERTIFICATE:
-            return self._create_certificate(value)
+            return self._create_certificate()
```
rellerreller: I plan to pass you a value for this just like with public and private keys. The value will be a DER encoded X.509 certificate.

kmip/tests/core/secrets/test_certificate.py
```diff
@@ -0,0 +1,378 @@
...
+    """
+
+    def setUp(self):
+        super(TestCertificate, self).setUp()
+
+        self.certificate_type_a = None
+        self.certificate_type_b = CertificateTypeEnum.X_509
+
+        self.certificate_value_a = None
+        self.certificate_value_b = (
```
rellerreller: Could you add a comment saying how these bytes were generated or where they came from.

PeterHamilton: It's actually a little lower down, above the encodings. I can move it up here though.

kmip/core/factories/secrets.py
PeterHamilton: Yep, the factory here would just be an extra (unnecessary) step. You can literally do:

    Certificate(certificate_type=CertificateTypeEnum.X_509, certificate_value=b'\x00\x01\x02...')
to create the certificate. It builds everything internal automatically.

PeterHamilton: Oh, but the `store_secret` method in the `kmip_secret_store` in Barbican currently only goes through the `SecretFactory` to create secrets. Are you going to branch out to handle certificates differently? If you're going to use `store_secret` for certificates, you'll need to change things around. Certificates don't have algorithms or cryptographic lengths.

rellerreller: OK, I can create the certificate object directly. In Barbican the API takes in optional parameters for algorithm and crypto lengths. I assume those are for the public key in the certificate. I'm not sure. I'll ignore those for now because Barbican Core will save them and just have the KMIP secret store store the certificate.

kmip/tests/core/secrets/test_certificate.py
rellerreller: I think moving it above would make it clearer.

PeterHamilton:
@rellerreller Ok, I shifted the comment up above the other encodings.

PR #13, opened 2015-05-14
Reorganizing test suite and updating tox configuration

PeterHamilton:
This change reorganizes the test suite, splitting it into two separate test suites, one for unit tests and one for integration tests. The tox configuration is also updated to reflect this change. Invoking tox now runs the unit tests by default. A separate tox environment is used to run the integration tests.


README.rst
```diff
@@ -138,10 +138,14 @@ To run the Python syntax and format compliance tests::
...
 
-To run the test suite against Python 2.7::
+To run the unit test suite against Python 2.7::
 
   $ tox -e py27
 
+To run the integration test suite against the default Python installation:
+
+  $ tox -e integration
+
```
rellerreller: Do you need a note about starting the KMIP server first?

PeterHamilton: That is an excellent point. Adding it now.

PeterHamilton: Actually, right now the integration tests automatically start up the PyKMIP server. We'll remove that once we put in the dynamic config for talking to (possibly) remote appliances.

I'll update the README with instructions on how to start the server once we make that change.

PR #14, opened 2015-05-20
Adding config argument to integration test suite

PeterHamilton:
This change adds a pytest config argument, allowing the integration test suite to be run with the client configured with a specific section from the PyKMIP configuration file. A simple integration test class is added to demonstrate how to use the client created using the config argument.

To run the integration test suite with a specific configuration, use:

$ tox -e integration -- --config <config-section>

where <config-section> is the name of the config file section to use for the PyKMIP client.

### DIFF
--- a/kmip/tests/integration/conftest.py
+++ b/kmip/tests/integration/conftest.py
@@ -1,0 +1,40 @@
+# Copyright (c) 2015 The Johns Hopkins University/Applied Physics Laboratory
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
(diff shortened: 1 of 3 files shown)


joel-coffman:
Rather than using separate sections of a configuration file, how difficult would it be to specify a different configuration file entirely?

I'm thinking that it would be best to distribute a "sample" config file with commented entries as part of PyKMIP or a config for a software server implementation. Co-mingling a single file with entries for different appliances would be more painful to version control and distribute because those credentials should not be shared.

PeterHamilton:
You mean splitting out different appliance config sections into their own individual config files? We certainly could go that route. I haven't fully thought through it yet though.

It would require reworking how the client handles config loading. Right now the client just expects the name of the config file section it should load settings from. We could change that to be a config file path. Or we could add a separate mechanism for specifying the location of the config file.

I actually have a partial patch now that improves how we load config settings, supporting multiple possible config file locations. Right now we just read from the config file included in the PyKMIP distribution; that's bad and really should only be used as a last resort option. Instead, we want to be able to read config from standard locations, like the user's home directory, ~/.config, or /etc. Placing appliance sensitive config information in config files in these locations would allow you to avoid mixing sensitive data with the repo.

I'm ok with exploring this route; however my short term goal here is to make sure that everything's in place by the end of this week for the integration test work. I don't want a rework of the config handling to block that.

joel-coffman:
> ...my short term goal here is to make sure that everything's in place by the end of this week for the integration test work.

I concur. This change would be nice to have but doesn't need to block this merge. I do want to continue to explore this route, as I think it will be very beneficial for testing in the long term.

PR #15, opened 2015-06-02
Updating client and demo logging

PeterHamilton:
This change updates various logging statements for the KMIP client and demo suites, ensuring that the default logging level for the library is at the INFO level and that all demo logging is at the INFO level.


rellerreller:
:+1: Looks good to me.

PR #16, opened 2015-06-02
Fixing infinite recursion bug with object inheritance

PeterHamilton:
This change removes all references to self.__class__. In object hierarchies with multiple levels of inheritance, using self.__class__ can cause an infinite loop when resolving references to parent classes.


rellerreller:
+1

PR #17, opened 2015-06-04
Removing old and out-of-date demo scripts

PeterHamilton:
This change removes the original top-level set of demo scripts in kmip/demos in favor of the functioning and up-to-date unit demo scripts located in kmip/demos/units.


rellerreller:
Looks good to me.

PR #19, opened 2015-06-05
Adding the pie package for the simplified PyKMIP API

PeterHamilton:
This change adds the pie package, which will be used for the simplified PyKMIP API, which will be officially known as the Pie API.


rellerreller:
It's as easy as pie to use. I love it!

:+1:

PR #20, opened 2015-06-08
Updating tox to specify Python version for integration tests

PeterHamilton:
This change updates the tox configuration for the integration test suite, specifying that the integration tests should be run with Python2.7.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -16,6 +16,7 @@ commands = flake8 kmip/
 [testenv:integration]
 # Note: This requires local or remote access to a KMIP appliance or service
 deps = {[testenv]deps}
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -16,6 +16,7 @@ commands = flake8 kmip/
 [testenv:integration]
 # Note: This requires local or remote access to a KMIP appliance or service
 deps = {[testenv]deps}
+basepython=python2.7
```
rellerreller: I think this is good for now since Barbican is using Python 2.7 for their functional tests. We will have to monitor this to see when we should change it.

rellerreller:
:+1:

PR #21, opened 2015-06-10
Adding the ManagedObject base class

PeterHamilton:
This change adds ManagedObject, the abstract base class of the simplified KMIP object hierarchy. It defines a set of attributes common to all KMIP objects and registers placeholder methods to be implemented by all subclasses. A basic test suite is included, using a dummy ManagedObject subclass for testing.

### DIFF
--- a/kmip/pie/objects.py
+++ b/kmip/pie/objects.py
@@ -1,0 +1,100 @@
+# Copyright (c) 2015 The Johns Hopkins University/Applied Physics Laboratory
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+from abc import ABCMeta
+from abc import abstractmethod
+
+from six import add_metaclass
+
+
+@add_metaclass(ABCMeta)
+class ManagedObject:
+    """
+    The abstract base object of the simplified KMIP object hierarchy.
+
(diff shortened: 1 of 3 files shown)


kmip/pie/objects.py
```diff
@@ -0,0 +1,100 @@
...
+        self._initial_date = None
+        self._last_change_date = None
+
+    @property
+    def object_type(self):
+        """
+        Accessor and property definition for the object_type attribute.
+
+        Returns:
+            None: if object_type is unset.
```
rellerreller: Can object type ever be none? It seems like it should never be.

kmip/pie/objects.py
```diff
@@ -0,0 +1,100 @@
...
+    def __repr__(self):
+        pass
+
+    @abstractmethod
+    def __str__(self):
+        pass
+
+    @abstractmethod
+    def __eq__(self, other):
+        pass
```
rellerreller: What do you think will define equals?

kmip/pie/objects.py
PeterHamilton: That is true for classes that can be instantiated, since each will correspond to an actual KMIP type. For abstract classes in the hierarchy that don't correspond to a type, the object type is by definition None.

kmip/pie/objects.py
PeterHamilton: I really haven't decided yet, which is one of the reasons I chose not to define it here yet. We could go strict and force equality across all attributes. Or we could focus on equality of just a core set, like value, object type, unique identifier, and a few others.

kmip/pie/objects.py
rellerreller: I would change this to state that is returns object_type. It should never be null. All subclasses must define this method, so it will never be null.

kmip/pie/objects.py
rellerreller: OK, we can leave it undefined for now. I brought it up with Joel. I'm not sure what to do either. I thought in the future we would store these objects in a database for the server version. If that is the case then may need special logic here. By special I mean it may need to abide by certain conventions like is the case for Hibernate objects.

Let's leave it undefined for now.

kmip/pie/objects.py
PeterHamilton: Fine by me, I'll update the patch.

PR #23, opened 2015-06-11
Adding version module to improve version handling

PeterHamilton:
This change adds a top-level version module to hold the version definition for the library. This is then used in both kmip/__init__.py and setup.py, allowing version inspection on import of the library and automatic syncing of the version across the project.


rellerreller:
:+1:

PR #24, opened 2015-06-15
Enumerations are Unsigned Integers

rellerreller:
Enumerations are to be encoded as four byte unsigned integers. They
were previously being sent as signed integers. This caused failures
when trying to transmit values that are extensions. Extensions contain
the value 8 hex in the first nibble of the first byte.


PeterHamilton:
Do you want the test as a separate patch after I merge this one?

PR #27, opened 2015-06-17
Adding additional cryptographic base classes

PeterHamilton:
This change adds two additional abstract base classes for the Pie ManagedObject object hierarchy: CryptographicObject and Key. Test suites are provided for each base class, using a dummy subclass of each for testing. The object test suite package structure is also modified to collect all ManagedObject subclass tests in one package.


PeterHamilton:
**DO NOT AUTO MERGE THIS PATCH** It's dependent upon https://github.com/OpenKMIP/PyKMIP/pull/21 - once that's merged in, I'll update this patch.

PR #32, opened 2015-06-22
Fixing startup logic

callidus:
Removed "exec(open('kmip/version.py').read())" it is unclear what
this was attempting to accomplish but opening a Python file and
executing its content is analogous to the "import" statement.
This was also attempting to open and execute a file using a path
relative to the folder of invocation, this is a security flaw
even if it succeeds in finding such a file.

### DIFF
--- a/kmip/__init__.py
+++ b/kmip/__init__.py
@@ -17,7 +17,7 @@ import logging.config
 import os
 import sys
 
-exec(open('kmip/version.py').read())
+import version
 
 path = os.path.join(os.path.dirname(__file__), 'logconfig.ini')
 


callidus:
Please make this fix a priority, without this patch (or the preferred fix) I actually cant use the library at all since it tries to open a none-existent file.

PeterHamilton:
Yep, I should have a fix out today.

PeterHamilton:
Ok, I'm punting on `pbr` for now; I just don't have time to set it up the way I want it right now. I've replaced the relative path used with a lookup for the full path of `version.py`, so that should no longer break when using `kmip` in arbitrary directories. See ﻿https://github.com/OpenKMIP/PyKMIP/pull/35 for more information.

PeterHamilton:
I'm going to go ahead and close this request since https://github.com/OpenKMIP/PyKMIP/pull/35 should handle it. If this pops up again, definitely let me know (or file an issue).

PR #39, opened 2015-06-25
Fix a few bugs in Credential object

cactorium:
This was causing the PyKMIP server to fail silently when given valid requests with a Device Credential in the Request Header; maybe some logging should also be added?


PeterHamilton:
The server definitely needs a heavy overhaul. It's on the list but unfortunately the client and new API are taking precedence.

cactorium:
No problem, and no worries!

PR #41, opened 2015-06-26
Adding revoke operation

callidus:
Adding Revoke opperation

### DIFF
--- a/kmip/core/enums.py
+++ b/kmip/core/enums.py
@@ -620,3 +620,12 @@ class ObjectGroupMember(Enum):
 class StorageStatusMask(Enum):
     ONLINE_STORAGE      = 0x00000001
     ARCHIVAL_STORAGE    = 0x00000002
+
+class RevocationReasonCode(Enum):
+    UNSPECIFIED            = 0x00000001
+    KEY_COMPROMISE         = 0x00000002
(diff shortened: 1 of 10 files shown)


PeterHamilton:
I'm just curious: why do you need these operations and (maybe more importantly) how are you using them? I want to make sure I test them appropriately. Thanks!

callidus:
Ultimately I am working on key wrapping. I cant use a key on the HSM until it is in the active state, hence needing activate. It would seem that an activated key can then not be destroyed, so I needed the revoke operation to deactivate them prior to destruction.

callidus:
Setting the activation date attributes on a key will cause the server to automatically mark it active within a given window, but I needed something more add-hoc for my prepossess

callidus:
looks like I'll need to rebase this. I'll do it now

callidus:
rebased

PeterHamilton:
Yep, I can confirm that activating a secret blocks any destroy operation on it, at least from the client. Adding revoke definitely makes sense.

I'll try to review this one today as well. I never seem to have as much time as I think I will when I sit down to review things.

callidus:
I know that feeling all too well! Thanks for looking this over

PR #42, opened 2015-06-26
Adding a demo for the Activate operation

PeterHamilton:
This change adds a unit demo for the Activate operation. The operation accepts the UID of a KMIP managed object that needs to be activated for future use.


callidus:
Ah, I probably should have added demos with the operations, thanks for adding this.

PeterHamilton:
Not a problem, they're really easy to add at this point.

PR #47, opened 2015-07-01
Adding various date attribute values

callidus:
Adding various date attribute values


PeterHamilton:
:+1: Thanks! I'm definitely trying to move towards efficient use of the primitives, avoiding unnecessary subclassing. This patch is a good example of that.

callidus:
awesome :) thanks Peter

PR #50, opened 2015-07-03
Adding key-wrapping

callidus:
This will allow a key wrapping specification to be used in conjunction with a get operation to return a wrapped key.

### DIFF
--- a/kmip/tests/integration/services/test_integration.py
+++ b/kmip/tests/integration/services/test_integration.py
@@ -19,9 +19,11 @@ from testtools import TestCase
 from kmip.core.attributes import CryptographicAlgorithm
 from kmip.core.attributes import CryptographicLength
 from kmip.core.attributes import Name
+from kmip.core.attributes import UniqueIdentifier
 
 
 from kmip.core.enums import AttributeType
+from kmip.core.enums import BlockCipherMode
 from kmip.core.enums import CryptographicAlgorithm as CryptoAlgorithmEnum
 from kmip.core.enums import CryptographicUsageMask
 from kmip.core.enums import KeyFormatType as KeyFormatTypeEnum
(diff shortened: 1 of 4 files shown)


PeterHamilton:
And I just realized that you were probably referring to how your config details get loaded into the integration tests. We use a `pytest fixture` (see `kmip/tests/integration/conftest.py`) to dynamically create the client with the specified config and automatically attach that client to any test class that specifies that it needs the client.

All you should need to do, if you're creating a new test class, is provide the following annotation before the class definition:
`@pytest.mark.usefixtures("client")`
That should set things up properly.

You could also just add a new test case onto the `TestIntegration` test class in `kmip/tests/integration/services/test_integration.py`. It's up to you.

Let me know if you have any other questions.

callidus:
added tests, sorry for the delay on this, busy times.

PeterHamilton:
Not a problem, I'm swamped too. I'll try to review this soon.

kmip/tests/integration/services/test_integration.py
```diff
@@ -419,3 +487,75 @@ def test_symmetric_key_register_get_destroy(self):
...
+        self.assertEqual(dek_clear_result.uuid.value,
+                         dek_wrapped_result.uuid.value)
+
+        clear_key = (dek_clear_result.secret.key_block.key_value.
+                     key_material.value)
+
+        wrapped_key = (dek_wrapped_result.secret.key_block.key_value.value)
+
+        # the wrapped key data should not be the same as the clear key data
+        self.assertNotEqual(clear_key, wrapped_key)
```
rellerreller: This proves that the two keys are not equal, but I would like to see proof that the key was encrypted correctly. Can you add code to unencrypt the wrapped key and then compare to the clear key? This would prove to me that the key was successfully encrypted.

rellerreller:
This looks good overall. It was easy to read and nicely written. I had one comment on the test code.

kmip/tests/integration/services/test_integration.py
PeterHamilton: I'd recommend using `cryptography` to handle decrypting the key. We're eventually going to use it for a more robust software server.

PeterHamilton:
The key wrapping code in other parts of the library is finally getting a refresh and will be integrated across client and server.

PR #52, opened 2015-07-16
Adding customizable timeout support to the KMIP client

PeterHamilton:
This change adds support for a customizable timeout option for the KMIP client. The client will stop attempting connections or operations once the timeout is exceeded instead of waiting for the default system timeout. The default timeouts is 30 seconds.

### DIFF
--- a/kmip/core/config_helper.py
+++ b/kmip/core/config_helper.py
@@ -36,6 +36,9 @@ class ConfigHelper(object):
     DEFAULT_USERNAME = None
     DEFAULT_PASSWORD = None
 
+    # Timeout measured in seconds
+    DEFAULT_TIMEOUT = 30
+
     def __init__(self):
         self.logger = logging.getLogger(__name__)
 
@@ -78,6 +81,7 @@ class ConfigHelper(object):
                 return_value = default_value
                 self.logger.debug(DEFAULT_MSG.format(default_value,
                                                      config_option_name))
+        # TODO (peter-hamilton): Think about adding better value validation
         if return_value == self.NONE_VALUE:
             return None
         else:
--- a/kmip/services/kmip_client.py
+++ b/kmip/services/kmip_client.py
@@ -78,7 +78,7 @@ class KMIPProxy(KMIP):
                  cert_reqs=None, ssl_version=None, ca_certs=None,
                  do_handshake_on_connect=None,
                  suppress_ragged_eofs=None,
-                 username=None, password=None, config='client'):
+                 username=None, password=None, timeout=30, config='client'):
         super(KMIPProxy, self).__init__()
         self.logger = logging.getLogger(__name__)
         self.credential_factory = CredentialFactory()
@@ -87,7 +87,7 @@ class KMIPProxy(KMIP):
         self._set_variables(host, port, keyfile, certfile,
                             cert_reqs, ssl_version, ca_certs,
                             do_handshake_on_connect, suppress_ragged_eofs,
-                            username, password)
+                            username, password, timeout)
         self.batch_items = []
 
         self.conformance_clauses = [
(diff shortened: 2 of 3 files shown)


kmip/core/config_helper.py
```diff
@@ -35,6 +35,7 @@ class ConfigHelper(object):
     DEFAULT_SSL_VERSION = 'PROTOCOL_SSLv23'
     DEFAULT_USERNAME = None
     DEFAULT_PASSWORD = None
+    DEFAULT_TIMEOUT = 30
```
rellerreller: Can you add a comment about the units for this? I assume this is seconds. What about milliseconds?

kmip/services/kmip_client.py
```diff
@@ -859,3 +860,6 @@ def _set_variables(self, host, port, keyfile, certfile,
 
         self.password = conf.get_valid_value(
             password, self.config, 'password', conf.DEFAULT_PASSWORD)
+
+        self.timeout = conf.get_valid_value(
```
rellerreller: What are the checks to verify that a valid value is configured? What if a value is less than or equal to zero?

kmip/core/config_helper.py
PeterHamilton: Yep, I'll add one. It's seconds, `socket.settimeout` only accepts seconds as a float. I guess we could use milliseconds and manually convert up, if we really wanted to.

kmip/services/kmip_client.py
PeterHamilton: The timeout being equal to zero is fine, that's an accepted state by the socket; it effectively puts the socket into non-blocking mode. I'm not sure what would happen on negative values; I can add a check for that.

kmip/core/config_helper.py
rellerreller: Seconds is ok. I am ok with whatever as long as it is documented.

PeterHamilton:
The value validation should probably be pushed into the `ConfigHelper`; 'get_valid_value' should really be called 'get_value' since it doesn't really do any sanity or type checking. However, current validation is done in the client, so I stuck the timeout check there. Revamping how we handle configuration information is on my to-do list. This will include updating the `ConfigHelper` to do value checking.

kmip/services/kmip_client.py
```diff
@@ -217,6 +217,7 @@ def open(self):
             suppress_ragged_eofs=self.suppress_ragged_eofs)
         self.protocol = KMIPProtocol(self.socket)
 
+        self.socket.settimeout(self.timeout)
```
sedukull: Peter,

We have two notes to consider i believe, one is connection time out and another is socket time out. Connection time out especially should work when establishing a connection to kmip server. Another is socket timeout, used for kmip operations where kmip operations can perform especially batch operations EX: Kmip protocol when used for data encryption depending upon the batch operations supported and data payload, can still be a blocking socket call i believe. I mean once the connection is established, make the socket a blocking once again. Even if user provides timeout while creating a kmipproxy instance, he still with this implementation can provide only one timeout.

PeterHamilton: While I'm not opposed to the idea of better granularity when it comes to managing operation timeouts, I don't think that will work for this case. For the `socket` library, connection timeouts and socket timeouts are the same thing. Only one timeout value can be set, via `socket.settimeout(...)`, and it must be set before a connection is established to another socket. I do not believe that it can be reset once a connection is established; that would create unstable behavior. This timeout value applies to all subsequent socket operations, including `socket.connect`, `socket.sendall`, `socket.recv`, etc. (i.e., all `socket` operations used by the client).

Request messages in the KMIP 1.1 specification do not specify timeout limits for operations either. However, KMIP requests can specify the maximum response size (in bytes) that should be returned by the server (for large operation responses). It is also possible to execute an operation on a KMIP appliance as an asynchronous operation. In this case, the appliance should return an asynchronous correlation value (i.e., an operation token) that can be used to poll the appliance for the status of the operation. PyKMIP does not currently support either of these features.

sedukull: Regarding the unstable behaviour, iam not sure thats the case, the same socket can be set with a different timeout in a different context. KMIP as per protocol is TTLV and not a transport variant and so doesnt make sense for it to specify the timeout as per of specification. In anyways, i believe, that should not a big issue to deal with now.

kmip/services/kmip_client.py
```diff
@@ -217,6 +217,7 @@ def open(self):
...
         self.socket.connect((self.host, self.port))
```
sedukull: Do we want to catch a socket exception here, log and raise it. This helps two things i believe, one for audit in the log for connection issues, and one predictable exception code consumers can expect and abstracted from underlying socket codes.

PeterHamilton: Good point, I'll add in a catch here for `socket.timeout`.

rellerreller:
:+1:

PR #53, opened 2015-07-16
Adding a Pie ObjectFactory

PeterHamilton:
This change adds an ObjectFactory for the Pie object hierarchy. It provides easy translation to and from Pie objects to the original core object hierarchy. The factory is only used by the new Pie client and should never be used outside of this context. A unit test suite for the factory is included, as are small changes to the current Pie object hierarchy and corresponding test suites.

### DIFF
--- a/kmip/pie/factory.py
+++ b/kmip/pie/factory.py
@@ -1,0 +1,96 @@
+# Copyright (c) 2015 The Johns Hopkins University/Applied Physics Laboratory
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+from kmip.core import attributes
+from kmip.core import misc
+from kmip.core import objects as cobjects
+from kmip.core import secrets
+
+from kmip.pie import objects as pobjects
+
+
+class ObjectFactory:
+    """
+    A factory to convert between the Pie and core object hierarchies.
+    """
+
+    def __init__(self):
+        """
+        Construct an ObjectFactory.
+        """
+        pass
+
+    def convert(self, obj):
+        """
+        Convert a Pie object into a core secret object and vice versa.
+
+        Args:
+            obj (various): A Pie or core secret object to convert into the
+                opposite object space. Required.
+
+        Raises:
+            TypeError: if the object type is unrecognized or unsupported.
+        """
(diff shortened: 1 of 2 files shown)


rellerreller:
It looks like this is a couple of commits. One is for converting keys to be subclasses of Key, and the other is a factory. Is this true or do these all need to be submitted together?

PeterHamilton:
Yeah, this could be split. I originally included the object changes since it made testing the factory simpler.

rellerreller:
Let's split it and resubmit as two separate patches.

PeterHamilton:
Will do.

PeterHamilton:
**DO NOT AUTOMERGE:** This patch now includes a [supporting commit](https://github.com/OpenKMIP/PyKMIP/pull/55) updating the Pie key objects. Once the corresponding patch is merged, this one will be updated accordingly.

PeterHamilton:
**DO NOT AUTOMERGE:** This patch has been updated to reflect changes in its [supporting commit](https://github.com/OpenKMIP/PyKMIP/pull/55). Once the corresponding patch is merged, this one will be updated accordingly.

PeterHamilton:
Ok, this one's been rebased onto master and is ready for final review.

kmip/pie/factory.py
```diff
@@ -0,0 +1,96 @@
...
+        elif isinstance(obj, pobjects.PublicKey):
+            return self._build_core_key(obj, secrets.PublicKey)
+        elif isinstance(obj, secrets.PublicKey):
+            return self._build_pie_key(obj, pobjects.PublicKey)
+        elif isinstance(obj, pobjects.PrivateKey):
+            return self._build_core_key(obj, secrets.PrivateKey)
+        elif isinstance(obj, secrets.PrivateKey):
+            return self._build_pie_key(obj, pobjects.PrivateKey)
+        else:
+            raise TypeError("object type unsupported and cannot be converted")
```
rellerreller: The thing I do not like about this is the giant switch statement. I'm trying to think of alternatives.

rellerreller: What about having a method for each type of conversion? You would have convert_to_pie_symmetric_key(..), convert_to_kmip_symmetric_key(..), convert_to_pie_public_key(..), convert_to_kmip_public_key(..), etc.

I don't want to spend too much time on this, so let's have a decision by 1:00.

PeterHamilton: We could do that. I do prefer the current approach, since the interface to the factory is so simple, but if you think this is compelling we can change it. What makes you uncomfortable about the switch?

rellerreller: Any time I see type checks and giant if statements then it makes me feel uncomfortable. It makes me think that something is not right and makes me question extensibility.

PeterHamilton: I think if you want to use a factory in a generic way, you'll always need some sort of switch. If it's not inside the factory, it would have to be outside (again, if it's being used to handle generic input).

In light of this conversation, my recent update to https://github.com/OpenKMIP/PyKMIP/pull/54, specifically for the `register` operation, definitely leverages the generic-ness of `convert` here. I'll need to update the client if we want to change the factory.

rellerreller:
:+1:

Let's go ahead and merge this!

PR #54, opened 2015-07-16
Adding the Pie KmipClient

PeterHamilton:
This change adds the Pie ProxyKmipClient, which provides a simplified version of the original KMIP interface and which wraps the original client. The ProxyKmipClient supports basic CRUD operations for symmetric, public, and private keys. A unit test suite for the client is included.

### DIFF
--- a/kmip/pie/client.py
+++ b/kmip/pie/client.py
@@ -1,0 +1,366 @@
+# Copyright (c) 2015 The Johns Hopkins University/Applied Physics Laboratory
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import logging
+import six
+
+from kmip.core import enums
+from kmip.core import objects as cobjects
+
+from kmip.core.factories import attributes
+
+from kmip.pie import api
+from kmip.pie import exceptions
+from kmip.pie import factory
+from kmip.pie import objects as pobjects
+
+from kmip.services.kmip_client import KMIPProxy
+
+
+class ProxyKmipClient(api.KmipClient):
+    """
+    A simplified KMIP client for conducting KMIP operations.
+
+    The ProxyKmipClient is a simpler KMIP client supporting various KMIP
+    operations. It wraps the original KMIPProxy, reducing the boilerplate
+    needed to deploy PyKMIP in client applications. The underlying proxy
+    client is responsible for setting up the underlying socket connection
+    and for writing/reading data to/from the socket.
+
+    Like the KMIPProxy, the ProxyKmipClient is not thread-safe.
+    """
+    def __init__(self,
+                 hostname=None,
+                 port=None,
+                 cert=None,
+                 key=None,
+                 ca=None,
+                 ssl_version=None,
+                 username=None,
+                 password=None,
+                 config='client'):
+        """
+        Construct a ProxyKmipClient.
+
+        Args:
+            hostname (string): The host or IP address of a KMIP appliance.
+                Optional, defaults to None.
+            port (int): The port number used to establish a connection to a
+                KMIP appliance. Usually 5696 for KMIP applications. Optional,
+                defaults to None.
+            cert (string): The path to the client's certificate. Optional,
+                defaults to None.
+            key (string): The path to the key for the client's certificate.
+                Optional, defaults to None.
+            ca (string): The path to the CA certificate used to verify the
+                server's certificate. Optional, defaults to None.
+            ssl_version (string): The name of the ssl version to use for the
+                connection. Example: 'PROTOCOL_SSLv23'. Optional, defaults to
+                None.
+            username (string): The username of the KMIP appliance account to
+                use for operations. Optional, defaults to None.
+            password (string): The password of the KMIP appliance account to
+                use for operations. Optional, defaults to None.
+            config (string): The name of a section in the PyKMIP configuration
+                file. Use to load a specific set of configuration settings from
+                the configuration file, instead of specifying them manually.
+                Optional, defaults to the default client section, 'client'.
+        """
+        self.logger = logging.getLogger()
+
+        self.attribute_factory = attributes.AttributeFactory()
+        self.object_factory = factory.ObjectFactory()
+
+        # TODO (peter-hamilton) Consider adding validation checks for inputs.
+        self.proxy = KMIPProxy(
+            host=hostname,
+            port=port,
+            certfile=cert,
+            keyfile=key,
+            ca_certs=ca,
+            ssl_version=ssl_version,
+            username=username,
+            password=password,
+            config=config)
+
+        # TODO (peter-hamilton) Add a multiprocessing lock for synchronization.
+        self._is_open = False
+
+    def open(self):
+        """
+        Open the client connection.
+
+        Raises:
+            ClientConnectionFailure: if the client connection is already open
+            Exception: if an error occurs while trying to open the connection
+        """
+        if self._is_open:
+            raise exceptions.ClientConnectionFailure(
+                "client connection already open")
+        else:
+            try:
+                self.proxy.open()
+                self._is_open = True
+            except Exception as e:
+                self.logger.exception("could not open client connection", e)
+                raise e
+
+    def close(self):
+        """
+        Close the client connection.
+
+        Raises:
+            ClientConnectionNotOpen: if the client connection is not open
+            Exception: if an error occurs while trying to close the connection
+        """
+        if not self._is_open:
+            raise exceptions.ClientConnectionNotOpen()
+        else:
+            try:
+                self.proxy.close()
+                self._is_open = False
+            except Exception as e:
+                self.logger.exception("could not close client connection", e)
+                raise e
+
+    def create(self, algorithm, length):
+        """
+        Create a symmetric key on a KMIP appliance.
+
+        Args:
+            algorithm (CryptographicAlgorithm): An enumeration defining the
+                algorithm to use to generate the symmetric key.
+            length (int): The length in bits for the symmetric key.
+
+        Returns:
+            string: The uid of the newly created symmetric key.
+
+        Raises:
+            ClientConnectionNotOpen: if the client connection is unusable
+            KmipOperationFailure: if the operation result is a failure
+            TypeError: if the input arguments are invalid
+        """
+        # Check inputs
+        if not isinstance(algorithm, enums.CryptographicAlgorithm):
+            raise TypeError(
+                "algorithm must be a CryptographicAlgorithm enumeration")
+        elif not isinstance(length, six.integer_types) or length <= 0:
+            raise TypeError("length must be a positive integer")
+
+        # Verify that operations can be given at this time
+        if not self._is_open:
+            raise exceptions.ClientConnectionNotOpen()
+
+        # Create the template containing the attributes
+        attributes = self._build_key_attributes(algorithm, length)
+        template = cobjects.TemplateAttribute(attributes=attributes)
+
+        # Create the symmetric key and handle the results
+        result = self.proxy.create(enums.ObjectType.SYMMETRIC_KEY, template)
+
+        status = result.result_status.enum
+        if status == enums.ResultStatus.SUCCESS:
+            uid = result.uuid.value
+            return uid
+        else:
+            reason = result.result_reason.enum
+            message = result.result_message.value
+            raise exceptions.KmipOperationFailure(status, reason, message)
+
+    def create_key_pair(self, algorithm, length):
+        """
+        Create an asymmetric key pair on a KMIP appliance.
+
+        Args:
+            algorithm (CryptographicAlgorithm): An enumeration defining the
+                algorithm to use to generate the key pair.
+            length (int): The length in bits for the key pair.
+
+        Returns:
+            string: The uid of the newly created public key.
+            string: The uid of the newly created private key.
+
+        Raises:
+            ClientConnectionNotOpen: if the client connection is unusable
+            KmipOperationFailure: if the operation result is a failure
+            TypeError: if the input arguments are invalid
+        """
+        # Check inputs
+        if not isinstance(algorithm, enums.CryptographicAlgorithm):
+            raise TypeError(
+                "algorithm must be a CryptographicAlgorithm enumeration")
+        elif not isinstance(length, six.integer_types) or length <= 0:
+            raise TypeError("length must be a positive integer")
+
+        # Verify that operations can be given at this time
+        if not self._is_open:
+            raise exceptions.ClientConnectionNotOpen()
+
+        # Create the template containing the attributes
+        attributes = self._build_key_attributes(algorithm, length)
+        template = cobjects.CommonTemplateAttribute(attributes=attributes)
+
+        # Create the asymmetric key pair and handle the results
+        result = self.proxy.create_key_pair(common_template_attribute=template)
+
+        status = result.result_status.enum
+        if status == enums.ResultStatus.SUCCESS:
+            public_uid = result.public_key_uuid.value
+            private_uid = result.private_key_uuid.value
+            return public_uid, private_uid
+        else:
+            reason = result.result_reason.enum
+            message = result.result_message.value
+            raise exceptions.KmipOperationFailure(status, reason, message)
+
+    def register(self, managed_object):
+        """
+        Register a managed object with a KMIP appliance.
+
+        Args:
+            managed_object (ManagedObject): A managed object to register. An
+                instantiatable subclass of ManagedObject from the Pie API.
+
+        Returns:
+            string: The uid of the newly registered managed object.
+
+        Raises:
+            ClientConnectionNotOpen: if the client connection is unusable
+            KmipOperationFailure: if the operation result is a failure
+            TypeError: if the input argument is invalid
+        """
+        # Check input
+        if not isinstance(managed_object, pobjects.ManagedObject):
+            raise TypeError("managed object must be a Pie ManagedObject")
+
+        # Verify that operations can be given at this time
+        if not self._is_open:
+            raise exceptions.ClientConnectionNotOpen()
+
+        # Extract and create attributes
+        attributes = list()
+        if hasattr(managed_object, 'cryptographic_usage_masks'):
+            mask_attribute = self.attribute_factory.create_attribute(
+                enums.AttributeType.CRYPTOGRAPHIC_USAGE_MASK,
+                managed_object.cryptographic_usage_masks)
+
+            attributes.append(mask_attribute)
+
+        template = cobjects.TemplateAttribute(attributes=attributes)
+        object_type = managed_object.object_type
+
+        # Register the managed object and handle the results
+        secret = self.object_factory.convert(managed_object)
+        result = self.proxy.register(object_type, template, secret)
+
+        status = result.result_status.enum
+        if status == enums.ResultStatus.SUCCESS:
+            uid = result.uuid.value
+            return uid
+        else:
+            reason = result.result_reason.enum
+            message = result.result_message.value
+            raise exceptions.KmipOperationFailure(status, reason, message)
+
+    def get(self, uid):
+        """
+        Get a managed object from a KMIP appliance.
+
+        Args:
+            uid (string): The unique ID of the managed object to retrieve.
+
+        Returns:
+            ManagedObject: The retrieved managed object object.
+
+        Raises:
+            ClientConnectionNotOpen: if the client connection is unusable
+            KmipOperationFailure: if the operation result is a failure
+            TypeError: if the input argument is invalid
+        """
+        # Check input
+        if not isinstance(uid, six.string_types):
+            raise TypeError("uid must be a string")
+
+        # Verify that operations can be given at this time
+        if not self._is_open:
+            raise exceptions.ClientConnectionNotOpen()
+
+        # Get the managed object and handle the results
+        result = self.proxy.get(uid)
+
+        status = result.result_status.enum
+        if status == enums.ResultStatus.SUCCESS:
+            managed_object = self.object_factory.convert(result.secret)
+            return managed_object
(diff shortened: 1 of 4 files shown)


PeterHamilton:
**DO NOT AUTOMERGE:** This patch includes another patch adding in `ObjectFactory` support. Once that patch is merged, this one will be rebased to account for the update.

I just wanted to get this up quickly so it could be looked over.

PeterHamilton:
Updated patch to fix several faulty mock calls.

rellerreller:
This is a huge patch at over 2500 lines of code. We need to figure out a way to break this up into smaller patches before reviewing. It's too big at the moment.

kmip/pie/client.py
```diff
@@ -0,0 +1,340 @@
...
+
+    def open(self):
+        """
+        Open the client connection.
+
+        Raises:
+            ClientConnectionFailure: if the client connection is already open
+            Exception: if an error occurs while trying to open the connection
+        """
+        if self._is_open:
```
rellerreller: This does not seem thread safe. We should add a note at class level to indicate this. In the future if we want to add it then we can modify, but I do not have a requirement for this.

kmip/pie/client.py
```diff
@@ -0,0 +1,340 @@
...
+        # Create the algorithm, length, and usage mask attributes
+        algorithm_attribute = self.attribute_factory.create_attribute(
+            AttributeType.CRYPTOGRAPHIC_ALGORITHM,
+            algorithm)
+        length_attribute = self.attribute_factory.create_attribute(
+            AttributeType.CRYPTOGRAPHIC_LENGTH,
+            length)
+        mask_attribute = self.attribute_factory.create_attribute(
+            AttributeType.CRYPTOGRAPHIC_USAGE_MASK,
+            [CryptographicUsageMask.ENCRYPT, CryptographicUsageMask.DECRYPT])
```
rellerreller: I'm wondering what we should do about these flags. If they are applied to both then maybe we should omit. Plus not all asymmetric keys will be used for this. I'm tempted to leave empty and caller can change later, but what are your thoughts?

PeterHamilton:
Actually, this patch is only 1800 or so; you're looking at the included factory patch, which will be merged separately.

I can pull out the integration tests and the new unit demos and submit them separately. That will help.

kmip/pie/client.py
```diff
@@ -0,0 +1,340 @@
...
+        if status == ResultStatus.SUCCESS:
+            public_uid = result.public_key_uuid.value
+            private_uid = result.private_key_uuid.value
+            return public_uid, private_uid
+        else:
+            reason = result.result_reason.enum
+            message = result.result_message.value
+            raise KmipOperationFailure(status, reason, message)
+
+    def register_key(self, key):
```
rellerreller: Is this just for keys or certificates too?

kmip/pie/client.py
```diff
@@ -0,0 +1,340 @@
...
+        """
+        Register a symmetric or asymmetric key with a KMIP appliance.
+
+        Args:
+            key (various): A key to register. Usually a SymmetricKey,
```
rellerreller: Will they all have a common base class like ManagedObject or Key?

kmip/pie/client.py
```diff
@@ -0,0 +1,340 @@
...
+
+        status = result.result_status.enum
+        if status == ResultStatus.SUCCESS:
+            return
+        else:
+            reason = result.result_reason.enum
+            message = result.result_message.value
+            raise KmipOperationFailure(status, reason, message)
+
+    def __enter__(self):
```
rellerreller: For my own understanding, what do these do?

kmip/pie/client.py
PeterHamilton: Oooh, good point. I'll add in a `threading.Lock()`, that should fix that.

kmip/pie/client.py
PeterHamilton: They allow you to use the client as a context manager. You can now do this: `with KmipClient(...) as client:`. It automatically opens and closes the client for you.

kmip/pie/client.py
PeterHamilton: Right now it's just for keys. I was thinking about having a separate `register_certificate` method for certificates but I'm open to going back to the generic `register`. My thinking was that having an explicit method for keys would make it a little easier for developers to understand and use.

kmip/pie/client.py
PeterHamilton: Right now all keys should inherit from `Key`. I check for this right below here.

kmip/tests/unit/pie/test_client.py
```diff
@@ -0,0 +1,598 @@
...
+    @mock.patch('kmip.pie.client.KMIPProxy',
+                mock.MagicMock(spec_set=KMIPProxy))
+    def test_init(self):
+        """
+        Test that a KMIPClient can be constructed with valid arguments.
+        """
+        KmipClient(
+            hostname='127.0.0.1',
+            port=5696,
+            cert='/example/path/to/cert',
```
rellerreller: Seeing this here I'm wondering if we should add checks in constructor to validate that files exist before returning a new instance. This would alert up front about issue with files. Otherwise I imagine the caller will receive socket exception or something.

rellerreller: If we do want this then let's add a todo and do in the future.

kmip/pie/client.py
PeterHamilton: Hmm, I think I'm fine either way. My original motivation here was to simple mimic what we've done with the prior tests, since we know those work with our backend(s). If you want to pull these, we can.

kmip/tests/unit/pie/test_client.py
PeterHamilton: It would probably be an `ssl` error, since these values are passed directly to `ssl.wrap_socket`. Checking for existence would be prudent. I'll add the todo.

joel-coffman:
Is it possible to remove the demos module and point people toward the automated (integration) tests instead?

PeterHamilton:
@joel-coffman Regarding the demos, I'd really prefer to keep them. The integration tests are littered with asserts and value/type checks and are generally harder to parse through. The demos, especially these new simplified ones, do a really good job highlighting how easy it is to use the client. I don't want to lose that to the integration tests. Ideally I'd like to see us produce more documentation, leveraging the demo scripts, rather than cut code. Plus, I use the demos to do most of my own testing. :)

PeterHamilton:
@joel-coffman Regarding the base class, are we planning on multiple instantiations of it beyond the simplified `KmipClient`? I'm not opposed to defining one, but the original `KMIPProxy` client and the new `KmipClient` are very different in terms of their interface. I can change that of course, moving back to generic operation/method names for the interface, where each accepts `*args` and `**kwargs` to leave implementation and input details open, but that makes the new client less intuitive. I suppose we could compensate with documentation if we want to go that route.

Any client exposing more of the KMIP specification, like the `KMIPProxy` client, is going to require additional inputs beyond what the simplified client requires; there's no escaping that. The original plan was that we would limit the new `KmipClient` to just expose a subset of the KMIP specification focusing on core CRUD operations for the main object types, to facilitate ease-of-use. Anything more complex would be left to the original `KMIPProxy` client, and/or its replacement.

joel-coffman:
I'm okay leaving the demos, but there appears to be a lot of boilerplate code that could make future refactoring more difficult (if, for example, there's a desire to change the way that the demo scripts are executed).

In general, I'd like to think that the critical bits of the demo scripts could appear in docstrings rather than separate files. Or perhaps use the doctest format to provide both an example code snippet and the expected result.

joel-coffman:
The advantage that I see for an abstract base class is that we define that class as **the** interface. A single abstract base class also makes it easier to identify cases where the public API is changing. For example, the existing pull request makes the client's logger part of the public API (see `kmip/pie/client.py:72`). I'm not sure that's the intent but it's difficult to spot such cases in 300+ lines of code (and imagine if multiple people were adding new capabilities to the class!).

This issue isn't a huge deal in my opinion -- I'm just trying to plan ahead for when PyKMIP has lots of users and any unintentional changes to the API could cause significant problems.

PeterHamilton:
**DO NOT AUTOMERGE:** This patch has been updated to reflect changes in its [dependent patches](https://github.com/OpenKMIP/PyKMIP/pull/53). This patch will be split into multiple patches in an upcoming update.

PeterHamilton:
**DO NOT AUTOMERGE:** This patch has been split into several patches, specifically breaking out the integration test suite and demos from the original patch. New patches will be submitted for those components shortly. Further updates to this patch to reflect review comments are also incoming.

PeterHamilton:
@joel-coffman Re demos: we should definitely explore using docstrings and doctests to augment the demos. I actually looked into using doctests at one point but shied away from them; they appear to work well for simple tests but seem to breakdown quickly for anything of moderate complexity.

PeterHamilton:
@joel-coffman Re interface: Yeah, trying to anticipate how things will scale for usability is something I've been thinking about a lot lately. I agree with you, standardizing on an interface is ideal for long-term stability. I think working on the new `KmipClient` has helped (at least me) identify how to make things simpler. Ideally the interface would be trivially supported by the `KmipClient` but would also have space for the original full featured specification.

I'll think about this a bit more and get back to you.

PeterHamilton:
Ok, some things to note with the latest version of this patch:
1. I added basic synchronization to the client using a simple decorator (`def _synchronize`). However, I don't have a good way to test client sychronization at the moment (it turns out this is kinda hard to do in Python).
2. In light of @joel-coffman's suggestions about the interface, I've gone back to the original client operation method names (e.g., `create` instead of `create_symmetric_key`); I like these names more for a KMIP interface than the new ones I was using. Everything's been updated to match, so no worries there.
As always, once https://github.com/OpenKMIP/PyKMIP/pull/53 is merged in, I'll update this one.

kmip/pie/client.py
```diff
@@ -0,0 +1,380 @@
...
+            message = result.result_message.value
+            raise exceptions.KmipOperationFailure(status, reason, message)
+
+    @_synchronize
+    def register(self, secret):
+        """
+        Register a secret with a KMIP appliance.
+
+        Args:
+            secret (various): A secret to register. An instantiatable subclass
```
rellerreller: /(various)/(ManagedObject)/

kmip/pie/client.py
```diff
@@ -0,0 +1,380 @@
...
+from kmip.core.factories import attributes
+
+from kmip.pie import exceptions
+from kmip.pie import factory
+from kmip.pie import objects as pobjects
+
+from kmip.services.kmip_client import KMIPProxy
+
+
+class KmipClient:
```
rellerreller: I think an interface for this would be good. That way it makes it easier to extend.

kmip/pie/client.py
```diff
@@ -0,0 +1,380 @@
...
+    original KMIPProxy, reducing the boilerplate needed to deploy PyKMIP
+    in client applications.
+    """
+    def __init__(self,
+                 hostname='127.0.0.1',
+                 port=5696,
+                 cert=None,
+                 key=None,
+                 ca=None,
+                 ssl_version=ssl.PROTOCOL_SSLv23,
```
rellerreller: I think the ssl_version default should be better. I think this should be TLS 1.1 at a minimum and recommend 1.2.

kmip/pie/client.py
```diff
@@ -0,0 +1,380 @@
...
+            ca_certs=ca,
+            ssl_version=ssl_version,
+            username=username,
+            password=password,
+            config=config)
+
+        self._is_open = False
+        self._lock = multiprocessing.RLock()
+
+    def _synchronize(f):
```
rellerreller: I know I brought up multi-threading before, but I do not know that we need it. @joel-coffman what do you think about this?

kmip/pie/client.py
```diff
@@ -0,0 +1,380 @@
...
+            enums.AttributeType.CRYPTOGRAPHIC_LENGTH,
+            length)
+        mask_attribute = self.attribute_factory.create_attribute(
+            enums.AttributeType.CRYPTOGRAPHIC_USAGE_MASK,
+            [enums.CryptographicUsageMask.ENCRYPT,
+             enums.CryptographicUsageMask.DECRYPT])
+
+        # Create the template containing the attributes
+        attributes = [algorithm_attribute, length_attribute, mask_attribute]
+        template = cobjects.CommonTemplateAttribute(attributes=attributes)
```
rellerreller: Looks like same code as in create. Might want to make a helper method.

kmip/pie/client.py
```diff
@@ -0,0 +1,380 @@
...
+        result = self.proxy.register(object_type, template, secret)
+
+        status = result.result_status.enum
+        if status == enums.ResultStatus.SUCCESS:
+            uid = result.uuid.value
+            return uid
+        else:
+            reason = result.result_reason.enum
+            message = result.result_message.value
+            raise exceptions.KmipOperationFailure(status, reason, message)
```
rellerreller: This is common code that could also be put in a helper function.

kmip/pie/client.py
```diff
@@ -0,0 +1,380 @@
...
+
+    @_synchronize
+    def create_key_pair(self, algorithm, length):
+        """
+        Create an asymmetric key pair on a KMIP appliance.
+
+        Args:
+            algorithm (CryptographicAlgorithm): An enumeration defining the
+                algorithm to use to generate the key pair.
+            length (int): The length in bits for the key pair.
```
rellerreller: There is no documentation for the return value. Make sure to add a note that the UUID can be used for other operations.

README.rst
```diff
@@ -124,7 +124,7 @@ Server profiles:
 Testing
 -------
 The PyKMIP test suite is composed of two parts: a unit test suite composed of
-over 600 unit tests, and an integration test suite that runs various tests
+over 750 unit tests, and an integration test suite that runs various tests
```
rellerreller: nit: I would not document the number of unit tests. That will always be out of sync and difficult to maintain.

kmip/pie/client.py
PeterHamilton: TLS 1.1 should be fine. TLS 1.2 is only supported in Python 2.7.9 and 3.4.

kmip/pie/client.py
PeterHamilton: Oh, yeah, good catch. Will change.

kmip/pie/client.py
PeterHamilton: Ok, I'll go ahead and put together what I think will work for this.

kmip/pie/client.py
PeterHamilton: Yep, this could go into a `_build_key_attributes` method. I'll add that in.

kmip/pie/client.py
PeterHamilton: The return values for different types of operations will be different, which would make the code in the `if` branch unique for each operation (e.g., see `create_key_pair` where we return both key IDs). Is it worth it to make a separate method for the `else` block?

kmip/pie/client.py
PeterHamilton: Yeah, I forgot return value docs for these. I'll add them in.

kmip/pie/client.py
PeterHamilton: Correction: TLS 1.1 AND 1.2 are only supported in Python 2.7.9 and 3.4. TLS 1.0 is supported by all Python versions.

kmip/pie/client.py
joel-coffman: I think that multi-threading support is critical, but I agree that it might not be necessary for the initial implementation. (Adding synchronization barriers would **not** change the existing interface.)

If we're going to include synchronization, let's use an existing library (e.g., [oslo.concurrency.lockutils](http://docs.openstack.org/developer/oslo.concurrency/api/lockutils.html)) rather than implementing one ourselves. In the very least, the synchronization decorator belongs in a utils modules rather than here.

README.rst
PeterHamilton: That's a good point. I always update the README before a release; I'll make this change then. For now I'll simply remove this update and just leave it as is.

kmip/pie/client.py
PeterHamilton: I have a couple of questions.

1. I understand the desire to use an existing library for concurrency. If we get to the point where we need complex or robust concurrency handling, I agree that it would be useful. However, the decorator is only 4 lines of code and for now I think that'll be sufficient. Do we want to add another dependency just to replace the decorator?

2. We could move the decorator to a utils module. I put it in this class originally because it references the internal `_lock` attribute of the class. This is an implicit dependency for using this locking decorator; it will only work on `KmipClient` methods, or any other class that defines it's own lock. (As an aside, `oslo.concurrency.lockutils` uses a lock and semaphore pool instead, where you name your locks and specify them explicitly when using the decorator. Useful if your critical sections are accessed at various points in your architecture, maybe a little overkill here.)

Thoughts?

kmip/pie/client.py
rellerreller: Interesting. OK, let's make it TLS 1.0.

PeterHamilton: After testing this operationally, I'm actually going to remove all default values from the client constructor, aside from the config value. If a default is specified in the code, it will override the config file value which I've found in testing to be mildly unintuitive. Since a user may not see the default value in the client constructor call, but will likely see the value in the config file, if they differ and the client constructor default is used, it's not clear why the config file value isn't used. We can either update the basic config to use TLS 1.0 or change the way we dynamically load config values to make this more intuitive to users.

For now, I'm just making them all `None`.

PeterHamilton:
Cross posting this from https://github.com/OpenKMIP/PyKMIP/pull/60, since it's more appropriate here: 

Given the API interface is now called `KmipClient`, here are some alternative names for the actual `KmipClient`:

- `KmipClient` - we can leave it as is, since we can differentiate the interface and the class via imports
- `ProxyKmipClient` - designates that this client leverages the original proxy client underneath
- `PieKmipClient` - the first client representing the Pie API, seems odd to reference the package name in the client name but eh, whatever
- `BasicKmipClient` - designates that this client provides a simpler interface than the prior proxy client
- `SimpleKmipClient` - same as `BasicKmipClient` really, we considered this before but @rellerreller wasn't a fan
- `MinimalKmipClient` - designates that this client only provides basic CRUD operations for KMIP managed objects

I think users would be scared to use a `HazardousKmipClient`, just my $0.02.

Thoughts? Preferences? It's just a name find/replace to me. I kind of like `ProxyKmipClient` but `BasicKmipClient` and `MinimalKmipClient` would also work fine. Also tagging @joel-coffman in case he wants to chime in.

PeterHamilton:
@rellerreller @joel-coffman @kaitlin-farr Just a call for votes on the name for `KmipClient`. Feel free to propose alternatives as well.

kmip/pie/client.py
```diff
@@ -0,0 +1,389 @@
...
+            function: The original function protected by a synchronization
+                lock.
+        """
+        def wrapper(self, *args, **kwargs):
+            with self._lock:
+                return f(self, *args, **kwargs)
+
+        return wrapper
+
+    @_synchronize
```
rellerreller: @joel-coffman where do we stand on this? Do we even want to bother with synchronize?

kmip/pie/client.py
```diff
@@ -0,0 +1,389 @@
...
+        if not isinstance(managed_object, pobjects.ManagedObject):
+            raise TypeError("managed object must be a Pie ManagedObject")
+
+        # Verify that operations can be given at this time
+        if not self._is_open:
+            raise exceptions.ClientConnectionNotOpen()
+
+        # Extract and create attributes
+        attributes = list()
+        if hasattr(managed_object, 'cryptographic_usage_masks'):
```
rellerreller: This is ok for now. In the future it would be cooler to not have to check hasattr for each attribute for the request. We can brainstorm this after this has been merged.

kmip/pie/client.py
rellerreller: Thanks for pointing that out :)

kmip/pie/client.py
PeterHamilton: I absolutely agree.

kmip/tests/unit/pie/test_client.py
```diff
@@ -0,0 +1,598 @@
...
+            uuid=attr.UniqueIdentifier(
+                'aaaaaaaa-1111-2222-3333-ffffffffffff'))
+
+        with ProxyKmipClient() as client:
+            client.proxy.create.return_value = result
+
+            uid = client.create(algorithm, length)
+            client.proxy.create.assert_called_with(
+                enums.ObjectType.SYMMETRIC_KEY, template)
+            self.assertIsInstance(uid, six.string_types)
```
rellerreller: You could also check the return ID to make sure it equals the value from line 202.

kmip/tests/unit/pie/test_client.py
```diff
@@ -0,0 +1,598 @@
...
+                mock.MagicMock(spec_set=KMIPProxy))
+    def test_create_on_closed(self):
+        """
+        Test that a ClientConnectionNotOpen exception is raised when trying
+        to create a symmetric key on an unopened client connection.
+        """
+        client = ProxyKmipClient()
+        args = [enums.CryptographicAlgorithm.AES, 256]
+        self.assertRaises(
+            ClientConnectionNotOpen, client.create, *args)
```
rellerreller: Good test and nice that raised exception is self-explanatory.

kmip/pie/client.py
PeterHamilton: Just so everyone's on the same page, I left the synchronization code in (for now) because I still wanted feedback on my questions (way above) regarding using a separate `utils` module or `oslo.concurrency`, as @joel-coffman suggested. If we want to defer this conversation, I'm happy to pull the synchronization out.

rellerreller:
I'm good with it, but I would like @joel-coffman to comment on synchronize before merging.

rellerreller:
What is the plan for integration/functional tests?

kmip/tests/unit/pie/test_client.py
PeterHamilton: Good point. I'll do that.

PeterHamilton:
@rellerreller Regarding integration tests, see https://github.com/OpenKMIP/PyKMIP/pull/57.

PeterHamilton:
@rellerreller Are you ok with `ProxyKmipClient`? It's funny, I don't remember changing to that name last week but I obviously did. I'm happy to change it if you (or others) prefer something else.

PeterHamilton:
@rellerreller Regarding synchronization: yes, I definitely want @joel-coffman's feedback.

PeterHamilton:
@rellerreller Ok, I think this is good to go. If you see anything I missed, let me know.

rellerreller:
:+1: LGTM!

PR #55, opened 2015-07-21
Updating Pie API key objects

PeterHamilton:
This change makes some minor updates to the Pie key object hierarchy. It fixes the key subclasses to inherit from Key directly. It adds in support for the key format type attribute, which is required for low-level key encoding. It also improves repr and str functionality by using binascii.hexlify to represent the key values. The corresponding test suites are updated accordingly to reflect these changes.

### DIFF
--- a/kmip/pie/objects.py
+++ b/kmip/pie/objects.py
@@ -16,11 +16,10 @@
 from abc import ABCMeta
 from abc import abstractmethod
 
+import binascii
 import six
 
-from kmip.core.enums import CryptographicAlgorithm
-from kmip.core.enums import CryptographicUsageMask
-from kmip.core.enums import ObjectType
+from kmip.core import enums
 
 
 @six.add_metaclass(ABCMeta)
@@ -168,10 +167,12 @@ class Key(CryptographicObject):
         cryptographic_algorithm: A CryptographicAlgorithm enumeration defining
             the algorithm the key should be used with.
         cryptographic_length: An int defining the length of the key in bits.
+        key_format_type: A KeyFormatType enumeration defining the format of
+            the key value.
     """
 
     @abstractmethod
-    def __init__(self, value=None, algorithm=None, length=None):
+    def __init__(self):
         """
         Create a Key object.
         """
@@ -179,6 +180,7 @@ class Key(CryptographicObject):
 
         self.cryptographic_algorithm = None
         self.cryptographic_length = None
+        self.key_format_type = None
 
         # All remaining attributes are not considered part of the public API
         # and are subject to change.
@@ -201,6 +203,7 @@ class SymmetricKey(Key):
         cryptographic_algorithm: The type of algorithm for the SymmetricKey.
         cryptographic_length: The length in bits of the SymmetricKey value.
         value: The bytes of the SymmetricKey.
+        key_format_type: The format of the key value.
         cryptographic_usage_masks: The list of usage mask flags for
             SymmetricKey application.
         names: The string names of the SymmetricKey.
@@ -217,12 +220,14 @@ class SymmetricKey(Key):
             length(int): The length in bits of the key.
             value(bytes): The bytes representing the key.
             masks(list): A list of CryptographicUsageMask enumerations defining
-                how the key will be used.
-            name(string): The string name of the key.
+                how the key will be used. Optional, defaults to None.
+            name(string): The string name of the key. Optional, defaults to
+                'Symmetric Key'.
         """
         super(SymmetricKey, self).__init__()
 
-        self._object_type = ObjectType.SYMMETRIC_KEY
+        self._object_type = enums.ObjectType.SYMMETRIC_KEY
+        self.key_format_type = enums.KeyFormatType.RAW
 
         self.value = value
         self.cryptographic_algorithm = algorithm
@@ -255,7 +260,7 @@ class SymmetricKey(Key):
         if not isinstance(self.value, bytes):
             raise TypeError("key value must be bytes")
         elif not isinstance(self.cryptographic_algorithm,
-                            CryptographicAlgorithm):
+                            enums.CryptographicAlgorithm):
             raise TypeError("key algorithm must be a CryptographicAlgorithm "
                             "enumeration")
         elif not isinstance(self.cryptographic_length, six.integer_types):
@@ -266,7 +271,7 @@ class SymmetricKey(Key):
         mask_count = len(self.cryptographic_usage_masks)
         for i in range(mask_count):
             mask = self.cryptographic_usage_masks[i]
-            if not isinstance(mask, CryptographicUsageMask):
+            if not isinstance(mask, enums.CryptographicUsageMask):
                 position = "({0} in list)".format(i)
                 raise TypeError(
                     "key mask {0} must be a CryptographicUsageMask "
@@ -288,12 +293,12 @@ class SymmetricKey(Key):
     def __repr__(self):
         algorithm = "algorithm={0}".format(self.cryptographic_algorithm)
         length = "length={0}".format(self.cryptographic_length)
-        value = "value={0}".format(self.value)
+        value = "value={0}".format(binascii.hexlify(self.value))
 
         return "SymmetricKey({0}, {1}, {2})".format(algorithm, length, value)
 
     def __str__(self):
-        return str(self.value)
+        return str(binascii.hexlify(self.value))
 
     def __eq__(self, other):
         if isinstance(other, SymmetricKey):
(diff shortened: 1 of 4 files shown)


kmip/pie/objects.py
```diff
@@ -201,12 +206,13 @@ class SymmetricKey(Key):
...
         cryptographic_length: The length in bits of the SymmetricKey value.
         value: The bytes of the SymmetricKey.
+        format_type: The format of the key value.
         cryptographic_usage_masks: The list of usage mask flags for
             SymmetricKey application.
         names: The string names of the SymmetricKey.
     """
 
-    def __init__(self, algorithm, length, value, masks=None,
+    def __init__(self, algorithm, length, value, format_type, masks=None,
```
rellerreller: What values are acceptable for format_type? Would they all not be the same?

kmip/pie/objects.py
```diff
@@ -315,7 +331,7 @@ def __ne__(self, other):
             return NotImplemented
```
rellerreller: Why is this returning NotImplemented instead of False?

kmip/pie/objects.py
```diff
@@ -227,6 +235,7 @@ def __init__(self, algorithm, length, value, masks=None,
         self.value = value
```
rellerreller: Do we need a check to make sure values is actually a bytearray? I wish Python had type safety, so I do not know what is proper procedure.

kmip/pie/objects.py
```diff
@@ -342,6 +359,8 @@ def __init__(self, algorithm, length, value, masks=None,
                 type of algorithm for the key.
             length(int): The length in bits of the key.
             value(bytes): The bytes representing the key.
+            format_type(KeyFormatType): An enumeration defining the format of
+                the key value.
```
rellerreller: I think we should define what values are acceptable. I think public keys should be raw or X.509 and X.509 is the default. An X.509 means the value is a DER-encoded bytearray for a SubjectPublicKeyInfo ASN.1 structure. A raw type means the public key is a DER-encoded representation of the public key conforming to its ASN.1 spec. All other values should raise an exception.

kmip/pie/objects.py
```diff
@@ -353,6 +372,7 @@ def __init__(self, algorithm, length, value, masks=None,
         self.value = value
```
rellerreller: Add a TODO to verify format_type and value. I would like to see a check where if a user specifies X.509 as format type and value does not conform to that then an exception is raised.

kmip/pie/objects.py
```diff
@@ -407,17 +430,21 @@ def validate(self):
...
+        value = "value={0}".format(binascii.hexlify(self.value))
+        format_type = "format_type={0}".format(self.key_format_type)
 
-        return "PublicKey({0}, {1}, {2})".format(algorithm, length, value)
+        return "PublicKey({0}, {1}, {2}, {3})".format(
+            algorithm, length, value, format_type)
 
     def __str__(self):
-        return str(self.value)
+        return str(binascii.hexlify(self.value))
```
rellerreller: This is good for now, but what are your thoughts on possibly making this PEM format in future?

kmip/pie/objects.py
```diff
@@ -446,12 +473,13 @@ class PrivateKey(CryptographicObject):
...
         cryptographic_length: The length in bits of the PrivateKey.
         value: The bytes of the PrivateKey.
+        format_type: The format of the key value.
         cryptographic_usage_masks: The list of usage mask flags for PrivateKey
             application.
         names: The list of string names of the PrivateKey.
     """
 
-    def __init__(self, algorithm, length, value, masks=None,
+    def __init__(self, algorithm, length, value, format_type, masks=None,
```
rellerreller: Similar comment on format_type as for public keys and adding todo for checking format_type and value. I think raw, pkcs1, and pkcs8 for now. I know there is ECPrivateKey, but I don't know much about that one.

rellerreller:
I think it might be good to break out different managed objects from managed_objects.py. That file is getting quite large at 580 lines in this patch. There are still more objects to go with opaque, raw, certificate, and X509 certificate. I find a 1000 line file too difficult to read. How can we easily break this up?

kmip/pie/objects.py
PeterHamilton: That's a good point. I don't think anything else would apply to symmetric keys except for raw. Maybe opaque?

The only format type used with symmetric keys in the testing documentation is raw, so let's stick with that.

kmip/pie/objects.py
PeterHamilton: `NotImplemented` is a special singleton value in Python that indicates that the comparison operator does not know how to compare itself to the given object. It is **not** equivalent to the `NotImplemented` exception.

In Python, comparison operators work in series. For example, say you want to compare objects `a` and `b`, so you do `a == b`. However, `a` doesn't know how to compare itself to `b`, so it returns `NotImplemented`. Python then, dynamically, runs `b == a`. If `b` knows how to compare itself to `a`, then the result of that comparison is used. If `b`, like `a`, doesn't know how to compare itself to `a`, Python will default to a raw object identity comparison, defaulting to a memory address comparison of the two objects.

Note the truth of `a == b` does not necessarily imply that `a != b` is false. Additionally, the truth of `a == b` does not imply that `b == a` is true. These comparison operators can be defined uniquely for every object to evaluate however you see fit.

For more information, see the docs on the [NotImplemented singleton](https://docs.python.org/2/library/constants.html#NotImplemented) and [comparison operators](https://docs.python.org/2/reference/datamodel.html#object.__eq__).

kmip/pie/objects.py
PeterHamilton: Will do. How would we do such a check?

kmip/pie/objects.py
PeterHamilton: Ok, I think that's reasonable.

kmip/pie/objects.py
PeterHamilton: There are a ton of public/private key format types that we've never used. Pretty much all of them would not be supported here, since the key value is a byte string; the key value for the various transparent key formats are different types of structs. I think those three are reasonable for now.

kmip/pie/objects.py
PeterHamilton: I went with `binascii.hexlify` here because I just wanted a clear, numeric string representation of the bytes without any strange character displays. If keys are displayed or logged (rarely if ever, I know), would displaying in PEM format be appropriate?

kmip/pie/objects.py
PeterHamilton: It's generally not recommended to focus heavily on type checking in Python. As you are aware, Python relies heavily on [duck typing](https://en.wikipedia.org/wiki/Duck_typing#In_Python), where it's the interface of an object that matters, not its actual type. The onus is generally placed on good documentation, clear code, sound testing, and robust exception handling to make sure that code and data are used properly.

Given how everything is public and touchable in Python, type checking doesn't really guarantee much. A type check only verifies that things are valid at a specific point in time, at a specific point of execution. Anything before or after is vulnerable to modification.

I usually tend to shy away from lots of type checking, though for PyKMIP the pattern I've used is to define `validate` methods that check input data upon creation of new objects, like keys. In this case, I already check that `value` is a `bytes` object, down in the `validate` method below.

PeterHamilton:
I'm open to splitting it. There's a natural set of abstract and key objects, so I could see something like `pie/objects/abstract.py` and `pie/objects/keys.py`. Certificates could go in `pie/objects/certificates.py` and the other, miscellaneous data objects would go in `pie/objects/data.py` or even `pie/objects/misc.py`. Would that work?

kmip/pie/objects.py
PeterHamilton: We'll need to include pkcs1 here as well.

PeterHamilton:
@rellerreller Ok, I updated the default and accepted formats for the keys as we discussed. I haven't split `objects.py` into separate files yet - let me know what you think regarding my suggesting scheme above.

kmip/pie/objects.py
rellerreller: I agree with you on the type checking. Let's ignore that.

The checks in the future can use the cryptography library. We can load an object and if the bytes are not correct then cryptography will yell at us.

kmip/pie/objects.py
rellerreller: Wow a == b does not imply b == a? That seems wrong to me, but maybe someone thought about this longer than me. OK, let's do this the Python way.

kmip/pie/objects.py
rellerreller: I think it would make things easier to read and debug. This is ok. Don't change it.

kmip/pie/objects.py
rellerreller: True, add pkcs1 as well.

rellerreller:
It looks like you addressed all of the changes I requested. I am good with it. We can work on moving code around after this but before the next release.

rellerreller:
:+1:

PR #56, opened 2015-07-22
shutting down and closing the socket in destructor, helps in case clo…

sedukull:
…se is not called by consumers and other misc changes


PeterHamilton:
:+1: Thanks Santhosh!

PR #58, opened 2015-07-22
Adding unit demos for the Pie KmipClient

PeterHamilton:
This change adds a new suite of unit demos for the new Pie KmipClient. The demos show how to create the KmipClient and how to use all of its primary operations. Demo utilities are updated to support the new suite.


PeterHamilton:
**DO NOT AUTOMERGE:** This patch has a couple [supporting](https://github.com/OpenKMIP/PyKMIP/pull/53) [patches](https://github.com/OpenKMIP/PyKMIP/pull/54). Once its dependencies are merged, this patch will be updated accordingly.

PeterHamilton:
This is up-to-date and ready for final review.

kmip/demos/pie/create.py
```diff
@@ -0,0 +1,58 @@
...
+from kmip.core import enums
+from kmip.demos import utils
+
+from kmip.pie import client
+
+
+if __name__ == '__main__':
+    # Build and parse arguments
+    parser = utils.build_cli_parser(enums.Operation.CREATE)
+    opts, args = parser.parse_args(sys.argv[1:])
```
rellerreller: We should look at using cliff for this. I think that would be a useful library for some of this.

rellerreller:
:+1:

kmip/demos/pie/create.py
PeterHamilton: Sure, there are tons of options here.

PR #59, opened 2015-07-23
Remove AUTHORS.txt

### LINKED ISSUE #30: AUTHORS.txt is out of date
AUTHORS.txt has not been updated in 10 months. It should be updated or removed (as authors may be easily determined from Git's log).

joel-coffman:
The authors included in AUTHORS.txt hasn't been updated in nearly a
year and is outdated. Rather than tracking the authors explicitly, it
would be better to extract this information from Git's commit
history.

The following command can be used to retrieve a list of the authors
for all the patches contributed to PyKMIP:
    git log | grep ^Author | sort | uniq

Resolves #30


PeterHamilton:
:+1: I'm ok with this.

PR #60, opened 2015-07-24
Adding Pie client API

PeterHamilton:
This change adds an abstract base class defining the client interface for all future PyKMIP clients. Required arguments for different operations are explicitly specified, while args and kwargs are provided for optional arguments and additional feature support.

### DIFF
--- a/kmip/pie/api.py
+++ b/kmip/pie/api.py
@@ -1,0 +1,81 @@
+# Copyright (c) 2015 The Johns Hopkins University/Applied Physics Laboratory
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import abc
+import six
+
+
+@six.add_metaclass(abc.ABCMeta)
+class KmipClient:
+    """
+    A simplified KMIP client interface for conducting KMIP operations.
+
+    The KmipClient provides a simple external interface for various KMIP
+    operations and composes the bulk of the PyKMIP Pie API.
+    """
+    @abc.abstractmethod
+    def create(self, algorithm, length):
+        """
+        Create a symmetric key on a KMIP appliance.
+
+        Args:
+            algorithm (CryptographicAlgorithm): An enumeration defining the
+                algorithm to use to generate the symmetric key.
+            length (int): The length in bits for the symmetric key.
+        """
+        pass
+
+    @abc.abstractmethod
+    def create_key_pair(self, algorithm, length):
+        """
+        Create an asymmetric key pair on a KMIP appliance.
+
+        Args:
+            algorithm (CryptographicAlgorithm): An enumeration defining the
+                algorithm to use to generate the key pair.
+            length (int): The length in bits for the key pair.
+        """
+        pass
+
+    @abc.abstractmethod
+    def register(self, managed_object):
+        """
+        Register a managed object with a KMIP appliance.
+
+        Args:
+            managed_object (ManagedObject): A managed object to register. An
+                instantiatable subclass of ManagedObject from the Pie API.
+        """
+        pass
+
+    @abc.abstractmethod
+    def get(self, uid):
+        """
+        Get a managed object from a KMIP appliance.
+
+        Args:
+            uid (string): The unique ID of the managed object to retrieve.
+        """
+        pass
+
+    @abc.abstractmethod
+    def destroy(self, uid):
+        """
+        Destroy a managed object stored by a KMIP appliance.
+
+        Args:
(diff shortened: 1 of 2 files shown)


kmip/pie/api.py
```diff
@@ -0,0 +1,95 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import abc
+import six
+
+
+@six.add_metaclass(abc.ABCMeta)
+class KmipClientAPI:
```
rellerreller: /KmipClientAPI/KmipClient/

kmip/pie/api.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    A simplified KMIP client interface for conducting KMIP operations.
+
+    The KmipClientAPI provides a simple external interface for various KMIP
+    operations and composes the bulk of the PyKMIP Pie API.
+    """
+    @abc.abstractmethod
+    def create(self, algorithm, length, *args, **kwargs):
```
rellerreller: Let's remove the args and kwargs. The problem is that if a caller is using a generic KmipClientAPI and passes those arguments then only certain instances can utilize them. This can cause inconsistent behavior if the KmipClient is swapped out with something else.

kmip/pie/api.py
```diff
@@ -0,0 +1,95 @@
...
+            length (int): The length in bits for the key pair.
+            args (various): Additional positional arguments for creating key
+                pairs.
+            kwargs (various): Additional keyword arguments for creating key
+                pairs.
+        """
+        pass
+
+    @abc.abstractmethod
+    def register(self, secret, *args, **kwargs):
```
rellerreller: nit: what about key or managed object instead of secret? You have 60 seconds to decide. Not a big deal for me if stays or changes.

kmip/pie/api.py
PeterHamilton: This will name clash with the actual `KmipClient`.

kmip/pie/api.py
PeterHamilton: I guess that's true. Without these though I don't see how we can have a generic interface that will satisfy both the simplified client and the whole spec.

kmip/pie/api.py
PeterHamilton: *panics* Umm, not `key` since `register` can be used for other objects.....I think I still prefer `secret`? But a `template` can be registered too and that's definitely not a `secret`. Ok, I'm fine with `managed_object`. Changing now...

PeterHamilton:
`tox` was whining about `pass` statements being untested so I threw in a short test suite. Nothing big.

Updated all references to `secret` to `managed_object`.

kmip/pie/api.py
rellerreller: I'm not a fan of *API, *Interface, I*, etc. class names. I feel like this is KmipClient and the existing KmipClient should be something like PieKmipClient or HazardousKmipClient to show that it maps from KmipClient interface to older KmipClientProxy.

kmip/pie/api.py
rellerreller: args and kwargs are still included.

PeterHamilton:
Ok, I pulled out `*args` and `**kwargs`. However, without those, how do we create clients that expose more of the KMIP interface? For example, say I create a new client and want the `get` operation to accept both a `uid` and a `format_type`. This interface doesn't represent that.

rellerreller:
> However, without those, how do we create clients that expose more of the KMIP interface?

You cannot. That is why the interface is so important. If we want to add more options in the future then we will have to modify the interface.

rellerreller:
:+1: Looks good!

kmip/pie/api.py
PeterHamilton: Ok, that makes sense. Here are alternative names for the actual `KmipClient`:

- `KmipClient` - we can leave it as is, since we can differentiate the interface and the class via imports
- `ProxyKmipClient` - designates that this client leverages the original proxy client underneath
- `PieKmipClient` - the first client representing the Pie API, seems odd to reference the package name in the client name but eh, whatever
- `BasicKmipClient` - designates that this client provides a simpler interface than the prior proxy client
- `MinimalKmipClient` - designates that this client only provides basic CRUD operations for KMIP managed objects

I think users would be scared to use a `HazardousKmipClient`, just my $0.02.

Thoughts? Preferences? It's just a name find/replace to me. I kind of like `ProxyKmipClient` but `BasicKmipClient` and `MinimalKmipClient` would also work fine. Also tagging @joel-coffman in case he wants to chime in.

kmip/pie/api.py
```diff
@@ -0,0 +1,81 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import abc
+import six
+
+
+@six.add_metaclass(abc.ABCMeta)
+class KmipClient:
```
kaitlin-farr: I'm not sure where the previous conversation thread went, but I was getting the emails for it.  How is changing the name of KMIPProxy to something like HazardousKmipClient going to work without breaking the Barbican plugin?

PeterHamilton: We won't be changing the name of `KMIPProxy`. That's static for now. We're trying to finalize the name of the new `KmipClient` (see https://github.com/OpenKMIP/PyKMIP/pull/54). It provides a simpler interface and wraps the original `KMIPProxy`.

PR #63, opened 2015-07-28
Adding certificates to the Pie object hierarchy

PeterHamilton:
This change adds certificate objects to the Pie object hierarchy, specifically a generic Certificate class and an X509Certificate subclass. A unit test suite covering both classes is included.


PeterHamilton:
Forgot to add some stuff to this one, hold off on reviewing until I push another update.

PeterHamilton:
Ok, this one should be ready to go now.

kmip/pie/objects.py
```diff
@@ -599,3 +599,158 @@ def __ne__(self, other):
...
+class Certificate(CryptographicObject):
+    """
+    The Certificate class of the simplified KMIP object hierarchy.
+
+    A Certificate is a core KMIP object that is the subject of key management
+    operations. For more information, see Section 2.2 of the KMIP 1.1
+    specification.
+
+    Attributes:
+        certificate_type: The type of the Certificate.
```
rellerreller: Can you add class types to this?

kmip/pie/objects.py
```diff
@@ -599,3 +599,158 @@ def __ne__(self, other):
             return not (self == other)
         else:
             return NotImplemented
+
+
+class Certificate(CryptographicObject):
```
rellerreller: Is there a way to mark this abstract or anything? We should add a note that this should not be invoked directly but use one of the subclasses.

rellerreller:
A couple of small comments to address and then you should be good to go.

kmip/pie/objects.py
PeterHamilton: Oh geeze, yeah, I should just use the `X509Certificate` here. I'll change this.

kmip/pie/objects.py
PeterHamilton: What do you mean?

PeterHamilton: Oh, I guess you mean subclasses? If so, yes, you can now.

PeterHamilton:
Ok, this has been updated. `Certificate` is now an uninstantiatable base class and all certificate operations should use `X509Certificate` instead.

kmip/pie/objects.py
rellerreller: I meant in the description put certificate_type (class type here): The type...

rellerreller: Like you did on line 717.

rellerreller:
One small comment about comments. After that is fixed then merge it.

kmip/pie/objects.py
rellerreller: Never mind. I see what you did. I'll file this under id10t.

rellerreller:
:+1: LGTM!

PR #66, opened 2015-07-29
Adding secret data to the Pie object hierarchy

PeterHamilton:
This change adds a SecretData class to the Pie object hierarchy. A unit test suite covering the class is provided, as are updates to the Pie object factory and associated tests.


rellerreller:
:+1:

PR #67, opened 2015-07-29
Adding secret data examples for the ProxyKmipClient

PeterHamilton:
This change adds two examples showing how to use secret data with the ProxyKmipClient. The first is a unit demo showing how to register a secret using the register operation of the ProxyKmipClient. The second is an integration test showing how to register, get, and destroy a secret using the ProxyKmipClient.


rellerreller:
:+1: LGTM!

PR #68, opened 2015-07-29
Adding opaque objects to the Pie object hierarchy

PeterHamilton:
This change adds an OpaqueObject class to the Pie object hierarchy. A unit test suite covering the class is included, as are updates to the Pie object factory and associated tests.

### DIFF
--- a/kmip/pie/objects.py
+++ b/kmip/pie/objects.py
@@ -859,3 +859,94 @@ class SecretData(CryptographicObject):
             return not (self == other)
         else:
             return NotImplemented
+
+
+class OpaqueObject(ManagedObject):
+    """
+    The OpaqueObject class of the simplified KMIP object hierarchy.
+
+    OpaqueObject is one of several ManagedObjects and is one of the core KMIP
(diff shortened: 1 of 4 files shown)


kmip/pie/objects.py
```diff
@@ -599,3 +599,94 @@ def __ne__(self, other):
...
+
+    OpaqueObject is one of several ManagedObjects and is one of the core KMIP
+    objects that are the subject of key management operations. For more
+    information, see Section 2.2 of the KMIP 1.1 specification.
+
+    Attributes:
+        opaque_type: The type of the opaque value.
+    """
+
+    def __init__(self, value, opaque_type, name='Opaque Object'):
```
rellerreller: I don't understand why KMIP requires opaque_type.

rellerreller:
:+1:

kmip/pie/objects.py
PeterHamilton: Yeah, it doesn't really make sense for opaque objects. It's opaque, it's just bytes, it shouldn't have a type.

PR #69, opened 2015-07-30
Adding opaque object examples for the ProxyKmipClient

PeterHamilton:
This change adds two examples showing how to use opaque objects with the ProxyKmipClient. The first is a unit demo showing how to register an opaque object using the register operation of the ProxyKmipClient. The second is an integration test showing how to register, get, and destroy an opaque object using the ProxyKmipClient.


rellerreller:
:+1: LGTM!

PR #72, opened 2015-08-04
Adding payloads for the GetAttributeList operation

PeterHamilton:
This change adds request and response payloads for the GetAttributeList operation. It updates the payload factories to support these new objects and updates and adds all associated or required test suites. A new exception is also included that is thrown by objects attempting to parse invalid KMIP encodings.

### DIFF
--- a/kmip/core/exceptions.py
+++ b/kmip/core/exceptions.py
@@ -1,0 +1,21 @@
+# Copyright (c) 2015 The Johns Hopkins University/Applied Physics Laboratory
+# All Rights Reserved.
+#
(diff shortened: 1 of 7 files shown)


rellerreller:
At some point we need to decide a code freeze date for next release. This will probably make it, but we need to start limiting patches soon.

PeterHamilton:
@rellerreller Yep, I agree. This doesn't even need to make it in for 0.4.0; I'm happy to hold it until after the release.

PeterHamilton:
This patch has been rebased and is ready for review.

joel-coffman:
Looks okay to me

PR #76, opened 2015-08-05
Updating support for the LongInteger primitive

PeterHamilton:
This change updates the LongInteger primitive, improving class documentation, adding standard Python operators, cleaning up the original implementation, and expanding the corresponding unit test suite to address the modifications.

PR #79, opened 2015-08-11
Updating support for the DateTime primitive

PeterHamilton:
This change updates the DateTime primitive, adding documentation and Python string operator support. A new unit test suite for the primitive is included.

### DIFF
--- a/kmip/core/primitives.py
+++ b/kmip/core/primitives.py
@@ -16,6 +16,7 @@
 import logging
 import six
 import sys
+import time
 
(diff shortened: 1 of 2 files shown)


PeterHamilton:
**DO NOT AUTOMERGE:** Once https://github.com/OpenKMIP/PyKMIP/pull/76 is merged this patch will be updated.

hadesto:
:-1: 

Missing an assert on one of the tests

PR #80, opened 2015-08-11
Updating support for the Interval primitive

PeterHamilton:
This change updates the Interval primitive, adding a fresh implementation and documentation. A new unit test suite for the primitive is included.

### DIFF
--- a/kmip/tests/unit/core/primitives/test_interval.py
+++ b/kmip/tests/unit/core/primitives/test_interval.py
@@ -15,35 +15,211 @@
 
 import testtools
 
+from kmip.core import exceptions
+from kmip.core import primitives
 from kmip.core import utils
 
 
 class TestInterval(testtools.TestCase):
+    """
+    Test suite for the Interval primitive.
+    """
 
     def setUp(self):
         super(TestInterval, self).setUp()
-        self.stream = utils.BytearrayStream()
+
+        # Encoding and value based on Section 9.1.2 of the KMIP 1.1
+        # specification.
+        self.value = 864000
+        self.encoding = (
+            b'\x42\x00\x00\x0A\x00\x00\x00\x04\x00\x0D\x2F\x00\x00\x00\x00'
+            b'\x00')
+        self.encoding_bad_length = (
+            b'\x42\x00\x00\x0A\x00\x00\x00\x05\x00\x0D\x2F\x00\x00\x00\x00'
+            b'\x00')
+        self.encoding_bad_padding = (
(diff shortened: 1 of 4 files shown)


PeterHamilton:
**DO NOT AUTOMERGE:** This patch will be updated once its dependencies have been merged.

kmip/tests/unit/core/primitives/test_interval.py
```diff
@@ -15,35 +15,211 @@
...
 
     def setUp(self):
         super(TestInterval, self).setUp()
-        self.stream = utils.BytearrayStream()
+
+        # Encoding and value taken from Section 9.1.2 of the KMIP 1.1
+        # specification.
+        self.value = 864000
+        self.encoding = (
+            b'\x42\x00\x00\x0A\x00\x00\x00\x04\x00\x0D\x2F\x00\x00\x00\x00'
```
hadesto: This value, according to the spec is actually incorrect. 

The value for "an interval, containing the value for 10 days" (line 1914) is actually:
42 00 20 0A 00 00 00 04 00 0D 2F 00 00 00 00 00

PeterHamilton: Good eye! You're seeing the discrepancy with the third byte I believe, \x20 (in the spec) vs \x00 (here in the test). This is actually ok.

For these example encodings, the spec actually gives everything the dummy tag (the first 3 bytes of the encoding) \x42\x00\x20 (see the leading text in Section 9.1.2 of the spec). This tag is actually a legitimate tag for a CompromiseDate object (see Table 213, specifically pg. 107 of the spec). However, I didn't want all of my default primitives to have the tag for a legitimate KMIP object; I want a neutral tag that can apply to anything. So instead, I made up a new default tag, \x42\x00\x00, which is specifically unused by the spec. By default, all PyKMIP primitives start with this default tag; it's up to their instantiations or subclasses to specify their real tags, for whatever type of KMIP objects they actually represent. As a result, all of the encodings I take from the spec for these primitive tests are tweaked to make sure their tags line up properly with what my primitives will actual encode to. Hence the discrepancy.

hadesto: This might be nit-picking, but for the sake of clarity I would then suggest we say "based on". Thanks for the info though and for pointing out where the tags are explained, I was looking through the spec to see if I could find where the sections were explained :)

PeterHamilton: Not a problem. This is exactly why I appreciate reviews; they keep me on my toes!

hadesto:
:+1: 

I made a quick suggestion on wording for the test_intervals.py file. See pull request #87 for this.

PR #84, opened 2015-08-20
Updating integration tests to ignore software server

PeterHamilton:
This change adds a new pytest marker, ignore. It is used to silently skip the software server integration test suite, which is now broken from a recent ssl update. A pytest ini configuration file is also added to register the new marker.


PeterHamilton:
If we could get this reviewed and merged today, that'd be good. I want to get those broken tests off the board as fast as possible.

joel-coffman:
:+1:

hadesto:
:+1:

PR #85, opened 2015-08-26
Updating support for the Enumeration primitive

PeterHamilton:
This change updates the Enumeration primitive, adding a fresh implementation and documentation. An updated unit test suite for the primitive is included. Numerous changes to Enumeration usage across the library are also included to comply with the updated implementation.

### DIFF
--- a/kmip/core/attributes.py
+++ b/kmip/core/attributes.py
@@ -65,10 +65,9 @@ class Name(Struct):
 
     class NameType(Enumeration):
 
-        ENUM_TYPE = enums.NameType
-
         def __init__(self, value=None):
-            super(Name.NameType, self).__init__(value, Tags.NAME_TYPE)
+            super(Name.NameType, self).__init__(
+                enums.NameType, value, Tags.NAME_TYPE)
 
     def __init__(self, name_value=None, name_type=None):
         super(Name, self).__init__(tag=Tags.NAME)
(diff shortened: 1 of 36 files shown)


PeterHamilton:
**DO NOT AUTOMERGE:** Once the dependent patches are merged, this patch will be updated.

kmip/core/attributes.py
```diff
@@ -154,20 +153,17 @@ def create(cls, name_value, name_type):
 # 3.3
```
hadesto: nit:
Instead of having only the chapter number for the kmip spec, we should specifically state "KMIP Spec 1.1". Even more so that there are different versions and we don't know how this might change in the future. I would suggest we say "KMIP Spec 1.1: 3.3"

hadesto: Given that there are multiple versions of the KMIP spec, I would suggest we specifically state what version of the spec this applies to, along with the chapter. Something along the lines of:

"KMIP Spec 1.1 - 3.3"

kmip/core/messages/payloads/get.py
```diff
@@ -130,7 +128,7 @@ def read(self, istream):
         self.object_type.read(tstream)
         self.unique_identifier.read(tstream)
 
-        secret_type = self.object_type.enum
```
hadesto: Why did we change the variable name from "enum" to "value"?

hadesto:
:+1:

kmip/core/messages/payloads/get.py
PeterHamilton: The short reason is to make Enumeration consistent with the other primitives.

The longer (but still short) reason is that Enumeration used to be a subclass of Integer. Since Integer already defined "value", Enumeration had to use a different attribute to hold the enumeration value. Now that Enumeration is its own stand-alone class, it can use "value" without colliding with anything.

This simple switch caused a buuuuuunch of cascading changes, but I think its better for the long term.

PR #86, opened 2015-08-28
Updating the set of supported configuration file locations

PeterHamilton:
This change adds support for multiple configuration file storage locations, including the user's home directory and /etc. The preferred configuration file name is now pykmip.conf. The original name, kmipconfig.ini, will continue to be supported for legacy installations.


rellerreller:
:+1:

PR #89, opened 2015-09-08
Deleting broken server script

PeterHamilton:
This change removes an old and out-of-date script used to run the PyKMIP software server.


joel-coffman:
:+1:

PR #90, opened 2015-09-09
Adding version requirements to requirements.txt

guoyr:
It looks like pykmip doesn't support older versions of six, it'd be nice to check the version number when installing


PeterHamilton:
:+1: Thanks for this guoyr!

PR #93, opened 2015-09-17
Updating version handling

PeterHamilton:
This change updates how the library version number is dynamically loaded. It removes the use of exec in favor of a regular expression.


PeterHamilton:
I always meant to get around to this. Bandit forced my hand :smile:

PR #108, opened 2015-12-10
Fix BatchErrorContinuation message initialization

spencerjackson:
Hi, since we've updated to 0.4.1, PyKMIP's server has been emitting some errors when we send it messages:

    kmip.services.server.kmip_server - ERROR - KMIPServer <class 'TypeError'> enumeration type <class 'kmip.core.messages.contents.BatchErrorContinuationOpt
ion'> must be of type EnumMeta

I traced it down and it looks like kmip.core.messages.contents.BatchErrorContinuationOpt is passing its parent class a reference to itself, instead of its counterpart from kmip.core.enums. I modified the call to more closely resemble the other message contents objects which pass enumerations in this file. This appears to solve the issue we've been experiencing.

### DIFF
--- a/kmip/core/messages/contents.py
+++ b/kmip/core/messages/contents.py
@@ -240,7 +240,7 @@ class BatchErrorContinuationOption(Enumeration):
 
     def __init__(self, value=None):
         super(BatchErrorContinuationOption, self).__init__(
-            BatchErrorContinuationOption, value,
+            enums.BatchErrorContinuationOption, value,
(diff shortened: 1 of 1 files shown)


PeterHamilton:
:+1: Thanks for this!

Just a heads up: There's going to be a lot of changes made to the server over the next couple of months. I'm reworking it to add functionality and to clean up the rough edges. If you're using it consistently, you may want to stay on 0.4.0 for the time being as there will likely be breaks and backwards incompatible changes coming in at some point.

PR #110, opened 2015-12-17
services/server: use EOF exception

viktorTarasov:
In KMIPProtocol use 'EOFError' exception when there are no more bytes to read from input stream.
This allows to distinguish the 'end of transaction' from 'unexpected length of data read' and to avoid the error log at the end of transaction.

### DIFF
--- a/kmip/services/server/kmip_protocol.py
+++ b/kmip/services/server/kmip_protocol.py
@@ -36,8 +36,15 @@ class KMIPProtocol(object):
             self.socket.sendall(sbuffer)
 
     def read(self):
-        header = self._recv_all(self.HEADER_SIZE)
+        try:
+            header = self._recv_all(self.HEADER_SIZE)
+        except RequestLengthMismatch as e:
+            if e.received == 0:
+                raise EOFError("No request to process")
+            else:
+                raise
         msg_size = unpack('!I', header[4:])[0]
+
         payload = self._recv_all(msg_size)
         data = BytearrayStream(header + payload)
         self.logger.debug('KMIPProtocol.read: {0}'.format(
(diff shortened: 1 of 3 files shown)


PeterHamilton:
I like improving the error handling for the socket data. See my comments above and this should be good to go.

viktorTarasov:
- implemented the _RequestLengthMismatch_ exception as you suggested.
- the _EOFException_ is raised only when reading _Request.HEADER_.
- added unit tests for kmip-protocol to pass the _travis coverage/coveralls_ tests

viktorTarasov:
tests are re-implemented with _mock_

kmip/services/server/kmip_protocol.py
```diff
@@ -63,3 +72,22 @@ class KMIPProtocolFactory(object):
 
     def getProtocol(self, socket):
         return KMIPProtocol(socket)
+
+
+class RequestLengthMismatch(Exception):
+    """
+    An exception raised for error when the request read from stream
```
joel-coffman: nit: "for error" is unnecessary in my opinion

kmip/services/server/kmip_server.py
```diff
@@ -53,6 +53,7 @@ def serve(self):
         self.socket.listen(0)
         while True:
             connection, address = self.socket.accept()
+            self.logger.info('New connection')
```
joel-coffman: If adding a logging statement, is it possible to include more detailed information about the connection that would be useful when debugging?

kmip/services/server/kmip_server.py
```diff
@@ -70,9 +71,13 @@ def serve(self):
             try:
                 while True:
                     self._processor.process(protocol, protocol)
+            except EOFError as e:
+                self.logger.warning("KMIPServer {0} {1}".format(type(e), e))
             except Exception as e:
                 self.logger.error('KMIPServer {0} {1}'.format(type(e), e))
-                connection.close()
+
+            connection.close()
```
joel-coffman: Matter of style I suppose, but I prefer for `connection.close()` to be part of a `finally` block.

viktorTarasov:
* fixed spelling of comments;
* _close.connection_ moved to _finaly_ block;
* _address_ printed in debug message for new connection

viktorTarasov:
little refactoring of tests

viktorTarasov:
rebased on actual _master_

PR #111, opened 2015-12-17
demos/discover-versions: add protocol-versions option

viktorTarasov:
Introduce 'protocol-versions' option for the 'discover-versions' demo.


PeterHamilton:
Please fix the pep8 errors. I'll test this once those are fixed.

viktorTarasov:
re-submitted with updated PR#112

PR #112, opened 2015-12-17
server: process 'discover-versions'

viktorTarasov:
Support for 'DISCOVER_VERSIONS' operation


PeterHamilton:
Looks like there are some pep8 errors. Please fix those first.

PeterHamilton:
Thanks for the update. I like where this is headed. However, upon further review I think this needs to be split into two patches, one for the changes to the demo and one for the changes to the server. They're two separate chunks.

Don't worry about the `coveralls` failure above; I don't expect you to test all of `server.py`. :)

viktorTarasov:
as suggested, will be splitted into two patchs

Issue #116, opened 2016-01-05
Drop support for Python 2.6


PeterHamilton:
The `ssl` module in Python 2.6 does not support TLS 1.2. Given that other prominent open source projects, like OpenStack and pyca/cryptography, are moving away from supporting 2.6, PyKMIP should move to deprecate 2.6 support in the next release, ultimately dropping 2.6 support down the road.

rellerreller:
:+1:

PR #118, opened 2016-01-14
Adding pending deprecation warning for Python2.6

PeterHamilton:
This change adds a simple warning that is triggered whenever Python 2.6 is used with PyKMIP. It simply advises the user to use a newer version of Python. For now, Python 2.6 can still be used with PyKMIP.

Partially addresses https://github.com/OpenKMIP/PyKMIP/issues/116.

### DIFF
--- a/kmip/__init__.py
+++ b/kmip/__init__.py
@@ -17,6 +17,7 @@ import logging.config
 import os
 import re
 import sys
+import warnings
 
 # Dynamically set __version__
 version_path = os.path.join(os.path.dirname(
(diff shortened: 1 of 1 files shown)


kmip/__init__.py
```diff
@@ -63,3 +64,11 @@
         logging.basicConfig()
 
 __all__ = ['core', 'demos', 'services']
+
+if sys.version_info[:2] == (2, 6):
+    warnings.simplefilter("always")
+    warnings.warn(
+        ("Please use a newer version of Python (2.7.9+ preferred). PyKMIP "
+         "support for Python 2.6 will be abandoned in the future."),
```
joel-coffman: s/abandoned/deprecated
*or*
...will no longer support Python 2.6 in the future

PR #120, opened 2016-01-14
Adding server.crypto package and initial API

PeterHamilton:
This changes adds the kmip.services.server.crypto package and the initial API for the CryptographicEngine hierarchy.

### DIFF
--- a/kmip/services/server/crypto/api.py
+++ b/kmip/services/server/crypto/api.py
@@ -1,0 +1,68 @@
+# Copyright (c) 2016 The Johns Hopkins University/Applied Physics Laboratory
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+from abc import ABCMeta
+from abc import abstractmethod
+
+import six
+
+
+@six.add_metaclass(ABCMeta)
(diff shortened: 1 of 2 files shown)


kmip/services/server/crypto/api.py
```diff
@@ -0,0 +1,54 @@
...
+    @abstractmethod
+    def create_symmetric_key(self, algorithm, length):
+        """
+        Create a symmetric key.
+
+        Args:
+            algorithm(CryptographicAlgorithm): An enumeration specifying the
+                algorithm for which the created key will be compliant.
+            length(int): The length of the key to be created. This value must
+                be compliant with the constraints of the provided algorithm.
```
rellerreller: What's the return value?

kmip/services/server/crypto/api.py
```diff
@@ -0,0 +1,54 @@
...
+    @abstractmethod
+    def create_asymmetric_key_pair(self, algorithm, length):
+        """
+        Create an asymmetric key pair.
+
+        Args:
+            algorithm(CryptographicAlgorithm): An enumeration specifying the
+                algorithm for which the created keys will be compliant.
+            length(int): The length of the keys to be created. This value must
+                be compliant with the constraints of the provided algorithm.
```
rellerreller: What's the return value?

rellerreller:
:-1:

Documentation missing the return value.

PeterHamilton:
Yep, I haven't defined them yet. I'm implementing the `CryptographyEngine` now and once I've flushed out the details of what `pyca/cryptography` actually gives me, I'll fill it in.

Right now, it's looking like `create_symmetric_key` will return raw bytes, while `create_asymmetric_key_pair` will return two sets of bytes, one for each key. I'm going to look at what our appliances use as defaults for byte encoding and format and use those here as well. From what I can tell, it looks like those only change when you retrieve the actual key bytes with `Get`.

PeterHamilton:
Correction: the key values themselves will be bytes but I'll likely need to output other parameters and attribute values with those bytes so that they can be used to build valid `KeyBlock` objects. Or perhaps the engine should just build those objects directly. As I said before, still in flux.

PR #125, opened 2016-02-11
Fixing bug with DateTime str test

### LINKED ISSUE #124: Time to string test fails
I'm not sure I understand how far this issue goes, but the date_time test_str test fails in some locations.

```
_StringException: Traceback (most recent call last):
  File "/home/viraptor/Projects/PyKMIP/kmip/tests/unit/core/primitives/test_date_time.py", line 97, in test_str
    self.assertEqual(s, str(date_time))
  File "/home/viraptor/Projects/PyKMIP/.tox/py27/local/lib/python2.7/site-packages/testtools/testcase.py", line 362, in assertEqual
    self.assertThat(observed, matcher, message)
  File "/home/viraptor/Projects/PyKMIP/.tox/py27/local/lib/python2.7/site-packages/testtools/testcase.py", line 447, in assertThat
    raise mismatch_error
testtools.matchers._impl.MismatchError: 'Tue Aug 11 09:18:55 2015' != 'Tue Aug 11 08:18:55 2015'
```

Due to different dates the summer time is active, it looks like the test will fail in some locations (Australia, VIC) but not others. I'm not sure what the solution is here, but I think this code shouldn't try to use/interpret local times at all.

For comparison: Australia/VIC:
```
>>> datetime.datetime.fromtimestamp(time.mktime((2015, 8, 11, 9, 18, 55, 1, 223, 1)))
datetime.datetime(2015, 8, 11, 8, 18, 55)
>>> datetime.datetime.fromtimestamp(time.mktime((2015, 8, 11, 9, 18, 55, 1, 223, 0)))
datetime.datetime(2015, 8, 11, 9, 18, 55)
```
UTC:
```
>>> datetime.datetime.fromtimestamp(time.mktime((2015, 8, 11, 9, 18, 55, 1, 223, 1)))
datetime.datetime(2015, 8, 11, 9, 18, 55)
>>> datetime.datetime.fromtimestamp(time.mktime((2015, 8, 11, 9, 18, 55, 1, 223, 0)))
datetime.datetime(2015, 8, 11, 9, 18, 55)
```

PeterHamilton:
An issue with local daylight savings time causes a comparison error in some locales when checking against a hard-coded expected time string. The hard-coded value has been replaced with a dynamic value that accounts for locale variances.

Fixes #124

### DIFF
--- a/kmip/tests/unit/core/primitives/test_date_time.py
+++ b/kmip/tests/unit/core/primitives/test_date_time.py
@@ -14,7 +14,6 @@
 # under the License.
 
 import testtools
-import time
 
 from kmip.core import primitives
 from kmip.core import utils
@@ -90,8 +89,7 @@ class TestDateTime(testtools.TestCase):
         Test that the string representation of a DateTime is formatted
         properly.
         """
-        t = (2015, 8, 11, 9, 18, 55, 1, 223, 1)
-        s = "Tue Aug 11 09:18:55 2015"
+        expected = 'Tue Aug 11 13:18:55 2015'
+        date_time = primitives.DateTime(1439299135)
 
-        date_time = primitives.DateTime(int(time.mktime(t)))
-        self.assertEqual(s, str(date_time))
+        self.assertEqual(expected, str(date_time))
--- a/kmip/core/primitives.py
+++ b/kmip/core/primitives.py
@@ -977,7 +977,7 @@ class DateTime(LongInteger):
         return "DateTime(value={0}, tag={1})".format(self.value, self.tag)
 
     def __str__(self):
-        return time.ctime(self.value)
+        return time.asctime(time.gmtime(self.value))
 
 
 class Interval(Base):


kmip/tests/unit/core/primitives/test_date_time.py
```diff
@@ -91,7 +91,7 @@ def test_str(self):
         properly.
         """
         t = (2015, 8, 11, 9, 18, 55, 1, 223, 1)
-        s = "Tue Aug 11 09:18:55 2015"
+        s = str(time.ctime(int(time.mktime(t))))
```
joel-coffman: This change effectively inlines the code used by DateTime to convert to a string (see [DateTime:\__str__](https://github.com/OpenKMIP/PyKMIP/blob/20dbad5/kmip/core/primitives.py#L955)). I'm not convinced that this test case is useful in this case.

joel-coffman:
:-1: Not convinced the existing test is useful as fixed.

PeterHamilton:
@joel-coffman Why not? The point of this test is to ensure that the time data survives getting passed through the `DateTime` object. Replicating the code path used to create the time string in the test case ensures that any breaking changes to how `DateTime` creates or handles time strings is detected, should the relevant `DateTime` code get changed in the future.

What alternative do you have in mind?

joel-coffman:
@PeterHamilton Not sure that I have a good alternative. My complaint is that the existing code is nothing more than a consistency check. It's almost like saying that the code is implemented correctly because the same code in a test case produces the same result. Not a very strong guarantee in my opinion.

Is it possible to incorporate locale information into the prior version of the test case?

PeterHamilton:
@joel-coffman I see your point. The current approach works but incorporating locale information would be a more robust solution.

I'll investigate what it would take to change or inspect the locale information.

PeterHamilton:
@joel-coffman I'm more and more coming around to your way of thinking. In fact, I'm now so far round the bend that I'm not even sure checking the actual value of the returned time string is useful. Checking that just a string is returned may be sufficient.

It is possible to modify the locale, but it's not thread-safe and it relies on the chosen locale being available on all possible platforms (no guarantees there, obviously). Another option is to move away from using `time.ctime` in `__str__` and use a more neutral construct, like `time.strftime`, but that is also dependent upon the locale. I think I want to avoid modifying the locale in individual test cases and test suites, and since this really only is an issue for this one test, I don't think making any structural changes is warranted.

Thoughts?

joel-coffman:
@PeterHamilton What does the KMIP specification say about locales?

Regardless, I concur that the existing test case should be removed *if* this patch is the only way to fix #124.

PeterHamilton:
@joel-coffman Here's the description of the `DateTime` primitive from the KMIP spec (1.1):

`
Date-Time values are POSIX Time values encoded as Long Integers. POSIX Time, as described in IEEE Standard 1003.1 [IEEE1003-1], is the number of seconds since the Epoch (1970 Jan 1, 00:00:00 UTC), not counting leap seconds.
`

The examples in the spec and testing documents all use GMT time examples. As I said previously, we could move from using `time.ctime` in `DateTime.__str__` to using a combination of `time.gmtime` and `time.strftime`. That should fix the time zone problem. However, locales bundle in language information, so any sort of hard coded time string in a test case will break if a different locale is used because the language of the time string may be different. Hence why I wanted to dynamically generate the time string value in the first place.

PeterHamilton:
@joel-coffman Also, this test case has to exist in some form to cover line 956 in `kmip/core/primitives.py`. No other test case tests converting `DateTime` to a string.

joel-coffman:
The excerpt from the KMIP specification suggests that the best change would be using `time.gmtime` (UTC) throughout.

PeterHamilton:
@joel-coffman Correct. `time.gmtime` does not produce a string, it produces a `struct_time`. That struct would need to be fed through `time.asctime` (preferable since it is not supposed to be affected by the locale) or `time.strftime` (which is affected by the locale) to get a human readable string. Using `time.asctime(time.gmtime(...))` should address the issue in #124 and keep us immune from locale language issues down the road, if I've read the `time` docs correctly.

If locale language persists as a problem though, we will have two options: (1) verify that `DateTime.__str__` produces a string but ignore its actual value; or (2) dynamically generate the expected time string like the current patch, taking into account the change from `time.ctime` to `time.asctime(time.gmtime(...))`.

joel-coffman:
Okay, sounds like `time.asctime(time.gmtime(...))` is the immediate path forward. Thanks for the detailed explanation!

PR #132, opened 2016-02-17
payload/locate: missing argument in 'read' calls

viktorTarasov:
missing:
* _stream_ argument in _read()_ calls
* 'object_group_member' in request payload


PeterHamilton:
:+1: LGTM! Thanks for catching those errors.

PR #133, opened 2016-02-18
attributes: 'eq' and 'repr' methods for Name attr.

viktorTarasov:
_eq_, _repr_ and _str_ methods for _Name_ and it's inner _NameValue_ and _NameType_ classes
It will facilitate the coming implementation of _locate_ in _mem-repo_

### DIFF
--- a/kmip/core/attributes.py
+++ b/kmip/core/attributes.py
@@ -63,11 +63,43 @@ class Name(Struct):
         def __init__(self, value=None):
             super(Name.NameValue, self).__init__(value, Tags.NAME_VALUE)
 
+        def __eq__(self, other):
+            if isinstance(other, Name.NameValue):
+                if self.value == other.value:
+                    return True
+                else:
+                    return False
+            else:
+                return NotImplemented
+
+        def __repr__(self):
+            return "{0}(value={1})".format(
+                    type(self).__name__, repr(self.value))
+
+        def __str__(self):
+            return "{0}".format(self.value)
+
     class NameType(Enumeration):
 
         def __init__(self, value=None):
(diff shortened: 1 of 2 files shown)


kmip/core/attributes.py
```diff
@@ -63,12 +63,35 @@ class NameValue(TextString):
         def __init__(self, value=None):
             super(Name.NameValue, self).__init__(value, Tags.NAME_VALUE)
 
+        def __eq__(self, other):
+            if isinstance(other, Name.NameValue):
+                if self.value == other.value:
+                    return True
+            return False
```
rellerreller: The other instances where __eq__ is implemented the final return statement returns NotImplemented for some reason. You would need to ask Peter why that is.

rellerreller:
:-1: 

There should be some unit tests to verify the implementation.

viktorTarasov:
fixing remarques from @rellerreller :
- _is-not-instance-of_ in '__eq__' returns _NotImplemented_;
- added unit tests for '__eq__' and '__str__' of _Name_, _NameType_ and _NameValue_.

PeterHamilton:
@viktorTarasov This patch looks good overall. I'll give it another pass on the next update.

viktorTarasov:
resolving comments of  @PeterHamilton:
- simplify return of _str_ magic method;
- no dummy test class to test _NotImplemented_ return.

viktorTarasov:
Sorry,
re-fixed remarques of @PeterHamilton:
-  magic _str_ of _Name_ attribute returns just a string;
-  unit test adjusted accordingly.

viktorTarasov:
rebased on actual _master_

PR #136, opened 2016-02-19
Added SQLAlchemy Code for Secret Data

rellerreller:
The SecretData class has the SQLAlchemy code in it to allow it be
stored in the database using the library.


PeterHamilton:
:+1: LGTM!

PR #137, opened 2016-02-19
Added Class Type to Managed Object

rellerreller:
SQLAlchemy requires a unique identifier for classes when using the
join table inheritence strategy. Previous commits were reusing the
object_type attribute of ManagedObject, but this strategy does not
work for classes that are subclasses of an object. This will be the
case for the Certificate object type because one such subclass will be
X509Certificate.


PeterHamilton:
:+1: LGTM!

PR #138, opened 2016-02-19
demos: new options to process by 'locate' demo

viktorTarasov:
options 'storage-status-mask' and 'object-group-member' are added to 'locate' demo


viktorTarasov:
Not actual

PR #140, opened 2016-02-22
Added New Managed Object Repository Interface

rellerreller:
Added a new interface for persisting managed objects. The new
interface provides methods for saving, retrieving, updating, and
deleting managed objects. It also provides methods for transaction
management.

The new interface is defined in the kmip.services.server.store module.
The module is not called repo because there is already a repo package.


PeterHamilton:
This looks good. My main question is: should there be a `search` method?

joel-coffman:
How practical is supporting transactions on a non-relational backend? Perhaps we should add those capabilities later when 1) they become necessary and 2) there are multiple implementations to demonstrate their feasibility?

rellerreller:
I think transactions on a non-relational backend are achievable. You could implement the command pattern and every operation is a new command that has an undo operation. In the event that you do a rollback then call undo on all of the commands. I don't think it would be too costly.

rellerreller:
Checkout the search method. There is a lot of documentation there.

One thing to note is that the attribute objects in the Pie API will likely need a flag to disable validation. The search method allows not all of the attributes of an attribute to be specified while doing a search.

kmip/pie/api.py
```diff
@@ -25,6 +25,51 @@ class KmipClient:
...
+    FRESH = "fresh"
+    LINK = "link"
+    APPLICATION_SPECIFIC_INFORMATION = "application_specific_information"
+    CONTACT_INFORMATION = "contact_information"
+    LAST_CHANGE_DATE = "last_change_date"
+    CUSTOM_ATTRIBUTE = "custom_attribute"
+    ALTERNATIVE_NAME = "alternative_name"
+    KEY_VALUE_PRESENT = "key_value_present"
+    KEY_VALUE_LOCATION = "key_value_location"
+    ORIGINAL_CREATION_DATE = "original_creation_date"
```
rellerreller: I added these here. We can reuse them when we define the locate function on the api.

kmip/services/server/store/repo.py
```diff
@@ -0,0 +1,194 @@
...
+        """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def search(self, attributes, max_items,
+               storage_status_mask=(enums.StorageStatusMask.ONLINE_STORAGE),
+               object_group_member=None):
+        """Search for a managed object.
+
+        Search for and return managed objects that have all of the attributes.
```
rellerreller: The locate operation only returns the UIDs of the managed objects. Do you want to just return UIDs or do you want the entire object? I was thinking that if I returned everything then the server can iterate over the results and pull out the UIDs if that is all it wants.

PeterHamilton: Hmmm, that's a good question. My main concern with returning the entire object is when a search is done with no filter. In that case, the data store should return *all* of the objects and if there are enough of them, that could be a problem.

joel-coffman:
:-1: I'm not comfortable trying to support transactions without putting considerable thought into the implementation.

*If* we're going to support concurrency at some point, then we definitely need synchronization between threads / processes.

> You could implement the command pattern and every operation is a new command that has an undo operation.

For the command pattern to "undo" an operation, the effect of that operation cannot be persistent until the transaction is committed. Otherwise, a crash leaves the data in an inconsistent state. But now we have to consider what happens when two threads try to access the same data simultaneously. Locking is one solution, but quite frankly, that's a lot of work to do correctly -- like reimplementing key portions of a database management system.

I think it's better not to promise anything about transactions and stick with the existing statement that the ManagedObjectRepo "is intended for single-threaded applications."

PeterHamilton:
For context, the `KmipEngine` (as it stands right now) enforces a lock on request processing, so only one batch of operations can ever be processed at a time. Since the `KmipEngine` wraps the data store, concurrency shouldn't be a problem here. It will still be important to note how the `ManagedObjectRepo` handles concurrency in the `ManagedObjectRepo` documentation.

rellerreller:
A couple of things on the transactions. The primary motivation for transactions is that they allow the server component to handle batches of commands. The KMIP spec defines many operations that can be performed, and it allows a client to perform multiple operations in a single request. This can be useful for doing an operation like _create_ a key and then add an _attribute_ to it.

The transaction is primarily needed so that if there are multiple steps then we can rollback any changes if any of the steps fail. I feel like this is high priority for the system.

If we want to support concurrency in the future then this also provides that by means of the database transactions.

> How practical is supporting transactions on a non-relational backend?

How many other backends do we really anticipate? I do not think creating a memory backend is worthwhile. You could do it, but why not use the DB repo with an in-memory database?

We could have a non-relational database, but why would anyone do that except for fun? If they did choose something else then I would hope that the database architecture supports transactions, and I'm sure that it would.

In summary I think if another backend is developed then I imagine it will not be a memory backend, and it will be some sort of non-relational database that has transaction support. I think the transaction support is crucial for batches of commands.

joel-coffman:
> How many other backends do we really anticipate? I do not think creating a memory backend is worthwhile. You could do it, but why not use the DB repo with an in-memory database?

If that's the case, why not skip this abstraction layer and use SQLAlchemy's facilities for transaction management directly?

rellerreller:
> If that's the case, why not skip this abstraction layer and use SQLAlchemy's facilities for transaction management directly?

Might as well add the abstraction, so that if we do want an alternative in the future then we have the capability.

PeterHamilton:
:-1: After discussing this, we'll stay with using `SQLAlchemy` directly and plan on adding in an explicit store object later, if needed.

PR #142, opened 2016-03-04
Adding the KmipEngine

PeterHamilton:
This change adds the KmipEngine, the core processing component of the KmipServer. The KmipEngine contains the KMIP application logic and handles process request messages by batch. The engine handles logging and error handling throughout the processing stack.

New server exceptions are added to handle new error cases. A test suite is included.

### DIFF
--- a/kmip/services/server/engine.py
+++ b/kmip/services/server/engine.py
@@ -1,0 +1,420 @@
+# Copyright (c) 2016 The Johns Hopkins University/Applied Physics Laboratory
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import logging
+import threading
+import time
+
+import kmip
+
+from kmip.core import enums
+from kmip.core import exceptions
+
+from kmip.core.messages import contents
+from kmip.core.messages import messages
+
+from kmip.core.messages.payloads import discover_versions
+from kmip.core.messages.payloads import query
+
+from kmip.core import misc
+
+from kmip.services.server.crypto import engine
+
+
+class KmipEngine(object):
+    """
+    A KMIP request processor that acts as the core of the KmipServer.
+
+    The KmipEngine contains the core application logic for the KmipServer.
+    It processes all KMIP requests and maintains consistent state across
+    client connections.
+
+    Features that are not supported:
+        * KMIP versions > 1.1
+        * Numerous operations, objects, and attributes.
+        * User authentication
+        * Batch processing options: UNDO
+        * Asynchronous operations
+    """
+
+    def __init__(self):
+        """
+        Create a KmipEngine.
+        """
+        self._logger = logging.getLogger(__name__)
+
+        self._cryptography_engine = engine.CryptographyEngine()
+        self._lock = threading.RLock()
+
+        self._id_placeholder = None
+
+        self._protocol_versions = [
+            contents.ProtocolVersion.create(1, 2),
+            contents.ProtocolVersion.create(1, 1),
+            contents.ProtocolVersion.create(1, 0)
+        ]
+
+        self._protocol_version = self._protocol_versions[0]
+
+    def _kmip_version_supported(supported):
+        def decorator(function):
+            def wrapper(self, *args, **kwargs):
+                if float(str(self._protocol_version)) < float(supported):
+                    name = function.__name__
+                    operation = ''.join(
+                        [x.capitalize() for x in name[9:].split('_')]
+                    )
+                    raise exceptions.OperationNotSupported(
+                        "{0} is not supported by KMIP {1}".format(
+                            operation,
+                            self._protocol_version
+                        )
+                    )
+                else:
+                    return function(self, *args, **kwargs)
+            return wrapper
+        return decorator
+
+    def _synchronize(function):
+        def decorator(self, *args, **kwargs):
+            with self._lock:
+                return function(self, *args, **kwargs)
+        return decorator
+
+    def _set_protocol_version(self, protocol_version):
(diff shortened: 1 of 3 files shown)


PeterHamilton:
This is the initial patch for the `KmipEngine`. It currently only supports `Query` and `DiscoverVersions`. I'll be iteratively adding functionality over the next couple of weeks, in addition to integrating it together with the rest of the `KmipServer` pieces.

kmip/services/server/engine.py
```diff
@@ -0,0 +1,404 @@
...
+
+from kmip.services.server.crypto import engine
+
+
+class KmipEngine(object):
+    """
+    A KMIP request processor that acts as the core of the KmipServer.
+
+    The KmipEngine contains the core application logic for the KmipServer.
+    It processes all KMIP requests and maintains consisten state across
```
rellerreller: /consisten/consistent/

kmip/services/server/engine.py
```diff
@@ -0,0 +1,404 @@
...
+    client connections.
+
+    Features that are not supported:
+        * KMIP versions > 1.1
```
rellerreller: At some point I would like to verify the 1.1 statement. My guess is that we can support 1.2 and 1.3 for the limited set of operations that we support.

kmip/services/server/engine.py
```diff
@@ -0,0 +1,404 @@
...
+                    "attack. {0}, {1}".format(then, now)
+                )
+                raise exceptions.InvalidMessage(
+                    "Stale request rejected by server."
+                )
+        else:
+            self._logger.info("Received request at time: {0}".format(
+                time.strftime(
+                    "%Y-%m-%d %H:%M:%S",
+                    time.gmtime(int(time.time()))
```
rellerreller: Why not use now?

kmip/services/server/engine.py
```diff
@@ -0,0 +1,404 @@
...
+        max_response_size = None
+        if header.maximum_response_size:
+            max_response_size = header.maximum_response_size.value
+
+        # Process the time stamp
+        if header.time_stamp:
+            then = header.time_stamp.value
+            now = int(time.time())
+
+            if (now >= then) and ((now - then) < 60):
```
rellerreller: I bet Joel will have a comment on this.

What if then is in the future?

kmip/services/server/engine.py
```diff
@@ -0,0 +1,404 @@
...
+
+    def _verify_credential(self, request_credential, connection_credential):
+        # TODO (peterhamilton) Add authentication support
+        # 1. If present, verify user ID of connection_credential is valid user.
+        # 2. If present, verify request_credential is valid credential.
+        # 3. If both present, verify that they are compliant with each other.
+        # 4. If neither present, set server to only allow Query operations.
+        pass
+
+    @_synchronize
```
rellerreller: Based on this annotation it appears that the intent is for only one request to be handled at a time? Is that correct?

It looks like your plan is to have multiple threads running for a thread-per-connection setup. While this will work it adds complexity and overhead. We could make this a single threaded application and do a reactor pattern. Essentially the for loop will be in the server. It will have the handle to the listener socket and each connected socket. Then all of the socket handles will be registered with an event dispatcher. When the event dispatcher returns then a handle to a socket is ready. If it is a connected socket then call process request. Otherwise create a client socket and register it with the event dispatcher.

See the reactor patter for more details.

I see that Twisted has a library for this.
http://twistedmatrix.com/documents/13.1.0/core/howto/reactor-basics.html

kmip/services/server/engine.py
PeterHamilton: Then `(now >= then)` would be false, triggering the `InvalidMessage` exception.

kmip/services/server/engine.py
PeterHamilton: I should. I added the time stamp now/then error checking after I added this log line.

kmip/services/server/engine.py
PeterHamilton: Yep, I saw this right after I posted the patch.

kmip/services/server/engine.py
PeterHamilton: I guess my thinking falls back to the features that distinguish the different versions. 1.2 adds a lot of new operations that we don't support and, in my mind, I don't want to claim 1.2 support if the bulk of that implies a lot of functionality that would be missing. Syntactically, you're right, we should be compatible. I just don't want to claim more than we should.

rellerreller: My argument would be that we claim 1.1 support, but we don't support lots of operations there as well.

kmip/services/server/engine.py
rellerreller: Ah, I see. Then the log message should state that invalid time. It could be past or future time.

kmip/services/server/engine.py
rellerreller: `
rlist = [server_socket]
while true:
  sockets_r_ready, , = select.select(rlist, None, None)
  for s in sockets_r_ready:
    if s == server_socket:
      rlist.append(s.connect())
    else:
      msg = read_message(s)
      process_request(msg)
      s.close()
      rlist.remove(s)
  `

PeterHamilton: Yes. For now, requests are handled one at a time, one per connection thread. In reality, this doesn't have to be the case. While a single `KmipEngine` will be shared across threads, synchronization only really has to protect data store operations (e.g., data creation, data persistence, data management). Multiple operations could be in execution simultaneously under this model. Some of the state-based features (e.g., the ID placeholder, the protocol version per request) could be persisted to the data store as well.

The reactor pattern could certainly work here. There are many options. I went with the thread-per-connection approach because it was the solution I knew I could get working fastest, since I've written this type of system before. To manage overhead, I'm capping the number of simultaneous connections to 5 (which is the recommended cap for `socket.listen`).

I had actually looked at Twisted several months ago when looking at Python server libraries. While powerful, at the time it looked like I would've had to figure out how to couple `kmip.core` encoding/decoding to a Twisted `Protocol` class, in addition to teasing out the SSL/TLS micromanagement that KMIP requires. It seemed like it would be hard to get working right. I'm sure I could've figured it out, but that was my thinking at the time.

kmip/services/server/engine.py
PeterHamilton: Agreed. I'll update the log messages accordingly.

kmip/services/server/engine.py
PeterHamilton: True! I'm fine with bumping up the supported versions.

PR #146, opened 2016-03-18
 Adding KmipEngine support for Register

PeterHamilton:
This change adds support for the Register operation to the KmipEngine. New exceptions and test cases are included.

### DIFF
--- a/kmip/core/messages/contents.py
+++ b/kmip/core/messages/contents.py
@@ -118,6 +118,58 @@ class ProtocolVersion(Struct):
         else:
             return NotImplemented
 
+    def __lt__(self, other):
+        if isinstance(other, ProtocolVersion):
+            if self.protocol_version_major < other.protocol_version_major:
+                return True
+            elif self.protocol_version_major > other.protocol_version_major:
+                return False
+            elif self.protocol_version_minor < other.protocol_version_minor:
+                return True
+            else:
+                return False
+        else:
+            return NotImplemented
+
+    def __gt__(self, other):
+        if isinstance(other, ProtocolVersion):
+            if self.protocol_version_major > other.protocol_version_major:
+                return True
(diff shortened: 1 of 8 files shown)


PeterHamilton:
This is currently dependent on #145.

kmip/core/messages/contents.py
```diff
@@ -118,6 +118,40 @@ def __ne__(self, other):
         else:
             return NotImplemented
 
+    def __lt__(self, other):
+        if isinstance(other, ProtocolVersion):
+            if self.protocol_version_major < other.protocol_version_major:
+                return True
+            elif self.protocol_version_minor < other.protocol_version_minor:
```
rellerreller: Do you need to add self.protocol_version_major == other.protocol_version_major to this elif statement? At this point we know that self.protocol_version_major >= other.protocol_version_major. What if self protocol version is 2.1 and other's is 1.2? I think this would report True for 2.1 < 1.2.

kmip/core/messages/contents.py
```diff
@@ -118,6 +118,40 @@ def __ne__(self, other):
...
+            elif self.protocol_version_minor > other.protocol_version_minor:
+                return True
+            else:
+                return False
+        else:
+            return NotImplemented
+
+    def __le__(self, other):
+        if isinstance(other, ProtocolVersion):
+            return self.__eq__(other) or self.__lt__(other)
```
rellerreller: These would probably be faster to write out code like above and change < to <=.

kmip/core/messages/contents.py
PeterHamilton: Oh wow, good catch! I can't believe I missed that. I'll fix this in dependent patch #145.

kmip/core/messages/contents.py
PeterHamilton: True. I can update these.

PR #147, opened 2016-03-21
 Adding KmipEngine support for Create

PeterHamilton:
This change adds support for the Create operation to the KmipEngine. New exceptions and test cases are included.


PeterHamilton:
This is dependent on #145 and #146.

PR #148, opened 2016-03-22
Adding KmipEngine support for CreateKeyPair

PeterHamilton:
This change adds support for the CreateKeyPair operation to the KmipEngine. New exceptions and test cases are included.


PeterHamilton:
This is dependent on #145, #146, and #147.

PR #149, opened 2016-03-24
fix: session: 'expected str instance, NoneType found'

viktorTarasov:
When KmipSession instantiated without session name there is error:
File ".../PyKMIP/kmip/services/server/session.py", line 57, in __init__
    self._logger = logging.getLogger('.'.join((__name__, name)))
TypeError: sequence item 1: expected str instance, NoneType found

### DIFF
--- a/kmip/services/server/session.py
+++ b/kmip/services/server/session.py
@@ -51,6 +51,9 @@ class KmipSession(threading.Thread):
             kwargs={}
         )
 
+        if name is None:
+            name = self.name
+
         self._logger = logging.getLogger('.'.join((__name__, name)))
 
         self._engine = engine
--- a/kmip/tests/unit/services/server/test_session.py
+++ b/kmip/tests/unit/services/server/test_session.py
@@ -38,6 +38,12 @@ class TestKmipSession(testtools.TestCase):
         """
         session.KmipSession(None, None, 'name')
 
+    def test_init_without_name(self):
+        """
+        Test that a KmipSession without 'name' can be created without errors.
+        """
+        session.KmipSession(None, None, None)
+
     def test_run(self):
         """
         Test that the message handling loop is handled properly on normal


kmip/services/server/session.py
```diff
@@ -51,6 +51,9 @@ def __init__(self, engine, connection, name=None):
             kwargs={}
         )
 
+        if name is None:
+            name = threading.current_thread().name
```
PeterHamilton: Good catch! However, this fix will pull the name from the parent thread (usually 'MainThread') and therefore allows for multiple threads to have the same name if created by the same parent thread. The `threading` module will auto name threads if `name=None` is passed to the `super(...).__init__(...)` call. Let's change the logging line below to use `self.name` instead of `name`; that should resolve the `NoneType` error.

kmip/tests/unit/services/server/test_session.py
```diff
@@ -38,6 +38,12 @@ def test_init(self):
         """
         session.KmipSession(None, None, 'name')
 
+    def test_init_without_name(self):
+        """
+        Test that a KmipSession without 'name' can be created without errors.
+        """
+        session.KmipSession(None, None, None)
```
PeterHamilton: Yep, should've included this when I first wrote this test suite. Thanks! :+1:

PeterHamilton:
This is good. Just make that small change and I'll merge it in.

kmip/services/server/session.py
viktorTarasov: Ok, I see, thanks.

By the way:
- does there any mailing list to ask the newbie questions? Like these ones:
- newbie questions:
  - what are/will you use for the threads? I'm trying to use the  SocketServer/AsynchronousMixins where KMIP session is run by it's RequestHandler. In this mode SocketServer makes thread per connection (with a distict thread name, but for a while I have not found the way to format this name). 
 - you've told _parent thread, MainThread_ -- does it mean that running server has to be separated  from the launching process into a dedicated thread? Can it run directly in the launching process?

PeterHamilton: Yes! There are a couple of places you can go to discuss PyKMIP. We have an IRC channel, #pykmip, on Freenode, hosted at https://webchat.freenode.net/. Login is free and registration is optional, if you haven't tried it before. We try to hang out there on the channel on a regular basis.

You can also send me or @rellerreller email with any questions you may have. This project isn't too active, and there aren't many people working on it, so we don't have a dedicated mailing list for it yet.

As for your questions, the `KmipSession` is a thread object that will be spun off from a new `KmipServer` class that I am finishing up right now; it's all part of the main KMIP server rewrite I've been working on for the past few months. The `KmipServer` handles receiving client connections and spins off `KmipSessions` for each one. The session then parses the incoming request and (if parse successful) passes the resulting KMIP request object to the `KmipEngine` for processing.

Since the `KmipServer` isn't out yet, there's no direct way to invoke `KmipSessions` yet without creating your own framework for it, which it sounds like you're doing with the SocketServer/AsynchronousMixins. It is definitely possible to run a `KmipSession` directly as your main thread; instead of invoking `KmipSession.start` (which kicks off the separate threaded-ness of the session) just call `KmipSession.run` directly.

The reason I wanted to avoid using the current thread's name is because the intent is to have multiple `KmipSessions` at once and I wanted to avoid any thread naming collisions. While I don't think there's anything technically incorrect about that thread naming scheme, thread activity is logged by thread name and I wanted to make it easy for those looking through the logs to distinguish between different threads.

Does that answer your questions? Definitely let me know if you have more. I'm happy to help.

viktorTarasov: Thank you very much for this developed answer, I'll be waiting for your _KmipServer_.

viktorTarasov:
Default session name is simplified, as has been suggested by @PeterHamilton.

PeterHamilton:
:+1: Looks good to me. Thanks @viktorTarasov!

PR #150, opened 2016-03-29
fix: kmip-client: batch item of error response

viktorTarasov:
Batch item of non-success response can be without operation attribute.


PeterHamilton:
:+1: Nice catch! I'm assuming you found this when testing against the new server? I saw errors due to this a few weeks back.

viktorTarasov:
Yes, I'm trying to use server and engines (with sqlalchemy behind).

PR #151, opened 2016-03-29
core messages: response batch item without 'operation'

viktorTarasov:
Batch item of error response message can be without 'operation'
attribute


PeterHamilton:
:+1: Once again, thanks for taking care of these.

PR #153, opened 2016-03-31
Linking the loggers for the different KmipServer components

PeterHamilton:
This change renames the loggers for different KmipServer components, ensuring that all server logs are processed by the same kmip.server logger.


PeterHamilton:
This is dependent on #152.

PR #154, opened 2016-03-31
Integrating the KmipEngine with the KmipSession

PeterHamilton:
This change adds KmipEngine support to the KmipSession. The session will now pass on received requests for processing by the engine, handling any results or errors that occur. The KmipSession test suite has been updated to reflect this change.


PeterHamilton:
This depends on #152 and #153.

PR #156, opened 2016-03-31
Removing default log configuration and usage

PeterHamilton:
This change removes the use of default logging settings in kmip.__init__.py as well as the bundled logconfig.ini file. Logging settings should be set by applications, not by underlying software libraries. All demos have been updated to set their own logging settings.


kmip/demos/pie/create_key_pair.py
```diff
@@ -23,6 +22,8 @@
 
 
 if __name__ == '__main__':
+    logger = utils.build_console_logger(logging.INFO)
```
rellerreller: How does this work for the user? Can they set the log level and logging file?

If not then the logger.debug statement will never be seen, and could be removed.

rellerreller:
One minor comment, but overall it looks good to me.

PeterHamilton:
@rellerreller Good point on the demo debug statements. I'll upgrade them all to `INFO` or `WARNING` as appropriate.

PR #157, opened 2016-03-31
Adding a server startup script

PeterHamilton:
This change adds a server startup script, bin/run_server.py, that can be used to run the KmipServer. Command-line arguments matching the server configuration file and KmipServer constructor are included.

### DIFF
--- a/bin/run_server.py
+++ b/bin/run_server.py
@@ -1,0 +1,156 @@
+#!/usr/bin/env python
+
+# Copyright (c) 2016 The Johns Hopkins University/Applied Physics Laboratory
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import optparse
+import sys
+
+from kmip.services import server
+
+
+def build_argument_parser():
+    parser = optparse.OptionParser(
+        usage="%prog [options]",
+        description="Run the PyKMIP software server.")
+
+    parser.add_option(
+        "-n",
+        "--hostname",
+        action="store",
+        type="str",
+        default="127.0.0.1",
+        dest="hostname",
+        help=(
+            "The host address the server will be bound to. A string "
+            "representing either a hostname in Internet domain notation or "
+            "an IPv4 address. Defaults to '127.0.0.1'."
+        ),
+    )
+    parser.add_option(
+        "-p",
+        "--port",
+        action="store",
+        type="int",
+        default=5696,
+        dest="port",
+        help=(
+            "The port number the server will be bound to. An integer "
+            "representing a port number. Recommended to be 5696 according to "
+            "the KMIP specification. Defaults to 5696."
+        ),
+    )
+    parser.add_option(
+        "-c",
+        "--certificate_path",
+        action="store",
+        type="str",
+        default=None,
+        dest="certificate_path",
+        help=(
+            "A string representing a path to a PEM-encoded server "
+            "certificate file. Defaults to None."
+        ),
+    )
+    parser.add_option(
+        "-k",
+        "--key_path",
+        action="store",
+        type="str",
+        default=None,
+        dest="key_path",
+        help=(
+            "A string representing a path to a PEM-encoded server "
+            "certificate key file. Defaults to None."
+        ),
+    )
+    parser.add_option(
+        "-a",
+        "--ca_path",
+        action="store",
+        type="str",
+        default=None,
+        dest="ca_path",
+        help=(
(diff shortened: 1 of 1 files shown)


bin/run_server.py
```diff
@@ -0,0 +1,154 @@
...
+    parser.add_option(
+        "-n",
+        "--hostname",
+        action="store",
+        type="str",
+        default=None,
+        dest="hostname",
+        help=(
+            "A string representing either a hostname in Internet domain "
+            "notation or an IPv4 address. Defaults to None."
```
rellerreller: Is this the network interface that will be listening for connections? What does None mean?

bin/run_server.py
```diff
@@ -0,0 +1,154 @@
...
+    )
+    parser.add_option(
+        "-p",
+        "--port",
+        action="store",
+        type="int",
+        default=None,
+        dest="port",
+        help=(
+            "An integer representing a port number. Recommended to be "
```
rellerreller: Port number to listen for connections? Again, why is None the default? I would have thought 5696 would be the default port number.

bin/run_server.py
```diff
@@ -0,0 +1,154 @@
...
+    parser.add_option(
+        "-k",
+        "--key_path",
+        action="store",
+        type="str",
+        default=None,
+        dest="key_path",
+        help=(
+            "A string representing a path to a PEM-encoded server "
+            "certificate key file. Defaults to None."
```
rellerreller: /certifcate/private/

This says it defaults to None, but I believe the server requires certificate and key.

bin/run_server.py
```diff
@@ -0,0 +1,154 @@
...
+    parser.add_option(
+        "-a",
+        "--ca_path",
+        action="store",
+        type="str",
+        default=None,
+        dest="ca_path",
+        help=(
+            "A string representing a path to a PEM-encoded certificate "
+            "authority certificate file. Defaults to None."
```
rellerreller: Can the file have more than one CA root certificate? My guess is that it is like SSL and can have multiple entries. Might be good to say the format of the file.

bin/run_server.py
```diff
@@ -0,0 +1,154 @@
...
+    )
+    parser.add_option(
+        "-l",
+        "--log_path",
+        action="store",
+        type="str",
+        default=None,
+        dest="log_path",
+        help=(
+            "A string representing a path to a log file. Defaults to None."
```
rellerreller: None of stdout.

kmip/services/server/server.py
```diff
@@ -0,0 +1,343 @@
...
+        to the specified log directory.
+
+        The main KmipEngine request processor is created here, along with all
+        information required to manage KMIP client connections and sessions.
+
+        Args:
+            hostname (string): The host address the server will be bound to
+                (e.g., '127.0.0.1'). Optional, defaults to None.
+            port (int): The port number the server will be bound to
+                (e.g., 5696). Optional, defaults to None.
```
rellerreller: Hostname and port descriptions are better here than above.

rellerreller:
Have you thought about adding an entry point entry for the server? I think that would be helpful. We could have pykmip for the client and pykmip-server for the server?

bin/run_server.py
PeterHamilton: By definition PEM-encoded CA certificate files can contain multiple CA root certificates and/or root certificate chains. See https://docs.python.org/2/library/ssl.html#ssl-certificates.

bin/run_server.py
PeterHamilton: `stdout` is not a supported option right now, as it is handled differently by the Python `logging` utility. The `KmipServer` only supports file-based logging at the moment.

PeterHamilton:
@rellerreller Re: entry points - what do you mean? Like a command-line hook for running the client or server?

PeterHamilton:
@rellerreller Re: script arguments - All of the arguments to `run_script.py` default to `None` because of how the Python command-line parser handles value parsing. If a default is specified here, there is no way for the user to disable that command-line argument and let a lower level value (like a default value defined by the `KmipServer` constructor, or a configuration setting defined by the configuration file) take precedence. I ran into this problem before with the client.

If a default is specified here and you do not want it to be used *and* you don't want to specify it yourself, you would have to explicitly say `-n None` (for a hostname value of "None") and then the script would have to define explicit corner case code to interpret this as an actual Python `None` value instead of the actual hostname "None". Just the explanation alone is complicated. Allowing `None` as the default makes the argument handling simpler and does what the user expects when they omit command-line arguments.

Re: argument help strings - I'll pull over more of the details provided in the `KmipServer` documentation. That should clarify some of the details for the different arguments. My original thought was to avoid duplication but now I don't think that really matters.

rellerreller:
Re: Re: script arguments - I see what you are saying from a coding perspective, but the user is not going to understand that. As an end user I want to know what the default will be. That information is very helpful. Otherwise I'm going to have to go through netstat commands to figure out the default values.

rellerreller:
Re: Re: entry points - An entry point allows the user to invoke a script without have to run `python -m my module.start`. You can define a python script and set an entry point for it.

For example, the files under ${barbican}/bin have entry points for them. Then after you install barbican pip install barbican you can simply type `barbican-manage` from the command line. It's neat.

PeterHamilton:
@rellerreller I added argument defaults where appropriate. Some, like the paths, are user/system specific. Let me know if you have any other comments.

Issue #107, opened 2015-12-09
Server Certificates are expired. 


cnsolis:
The 2 following sample certificates are expired:
PyKMIP/kmip/demos/certs/server.crt
PyKMIP//kmip/tests/integration/utils/certs/server.crt

I checked the expiration date on both and got the following output:
 
notAfter=Aug 19 04:20:33 2015 GMT

PeterHamilton:
Thanks for the reminder on this. I'll update them.

PeterHamilton:
#158 will remove one set of certificates. The other set is actually referenced by the client unit tests so I'll need to craft a more refined solution for those.

PR #158, opened 2016-04-04
Removing out-of-date certificate files

PeterHamilton:
This change removes the testing certificate files. They are expired.


PeterHamilton:
This partially addresses #107.

PR #164, opened 2016-04-08
server-engine: fix crypto-length in Create

viktorTarasov:
In request paylod to create symmetric key the
length of cryptographic algorithm is optional.

As Peter suggested, in this case the strongest allowed key size
is used.

### DIFF
--- a/kmip/services/server/crypto/engine.py
+++ b/kmip/services/server/crypto/engine.py
@@ -60,6 +60,8 @@ class CryptographyEngine(api.CryptographicEngine):
                 algorithm for which the created key will be compliant.
             length(int): The length of the key to be created. This value must
                 be compliant with the constraints of the provided algorithm.
+                Optional. If not specified, the key length is the maximal
+                allowed for a given cryptographic algorithm.
 
         Returns:
             dict: A dictionary containing the key data, with the following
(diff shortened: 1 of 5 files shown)


kmip/services/server/crypto/engine.py
```diff
@@ -86,7 +86,13 @@ def create_symmetric_key(self, algorithm, length):
 
         cryptography_algorithm = self._symmetric_key_algorithms.get(algorithm)
 
-        if length not in cryptography_algorithm.key_sizes:
+        if length is None:
+            # If cryptograohic length is not defined,
+            #   the strongest allowed key size is used.
+            sizes = list(cryptography_algorithm.key_sizes)
+            sizes.sort()
+            length = sizes[len(sizes) - 1]
```
PeterHamilton: To get the last item, you can just do:
    length = sizes[-1]

kmip/services/server/engine.py
```diff
@@ -671,16 +671,8 @@ def _process_create(self, payload):
...
 
         length = object_attributes.get('Cryptographic Length')
-        if length:
+        if length is not None:
             length = length.value
-        else:
-            # TODO (peterhamilton) The cryptographic length is technically not
-            # required per the spec. Update the CryptographyEngine to accept a
-            # None length, allowing it to pick the length dynamically. Default
-            # to the strongest key size allowed for the algorithm type.
```
PeterHamilton: Now that I think about it, the cryptography engine's `create_symmetric_key` method will need to also include the `key_length` in the result dict, since the length is needed for storing the whole managed object (line 699 below).

kmip/tests/unit/services/server/test_engine.py
```diff
@@ -1607,51 +1607,6 @@ def test_create_omitting_attributes(self):
...
-            regex,
-            e._process_create,
-            *args
-        )
-
-        e._logger.info.assert_any_call(
-            "Processing operation: Create"
-        )
-        e._logger.reset_mock()
-
```
PeterHamilton: This whole block needs to be replaced with a test that verifies that the strongest key size is picked when the key length is not specified.

viktorTarasov:
Sorry, I have not seen your comments and rebased the branch.
In this rebase I added crypto-length into response payload, if it was not present in request payload.

I will review the code, with your other comments in head, tomorrow ...
 ... and fix coveralls missings.

viktorTarasov:
replaced by #170

PR #165, opened 2016-04-11
Added KMIP Server Entry Point

rellerreller:
Added console_scripts to setup.py, and added an entry for
pykmip-server. The pykmip-server application will run the PyKMIP
server. The two command line options below will run the server.

python -m kmip.services.server.server
pykmip-server

After adding the pykmip-server entry point it seemed like
bin/run_server.py could be removed. It provided the same
functionality, but this allows the application to be run after doing a
pip install. Eliminating it also removes the need to update both files
when command line options are changed.


PeterHamilton:
:+1: LGTM!

PR #167, opened 2016-04-15
links for Managed Cryptographic Objects

viktorTarasov:
Add links to managed crypto objects.

Some unit tests are probably to be added.

### DIFF
--- a/kmip/core/attributes.py
+++ b/kmip/core/attributes.py
@@ -13,6 +13,8 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+import six
+
 from kmip.core import enums
 
 from kmip.core.enums import CertificateTypeEnum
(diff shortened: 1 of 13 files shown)


viktorTarasov:
Few fixes, more unit tests

PeterHamilton:
Thanks @viktorTarasov. Sorry for the delay in reviewing this. I've been out of town for a while but I'll try to review soon.

viktorTarasov:
@PeterHamilton Never mind, I know what it is to have little time.

viktorTarasov:
@PeterHamilton 
Question about links: How do you see the management of links for CryptoObjects?

afais:
- cert <-> public-key <-> private-key links has to be set by server, without client participation;
- it's simple for generated key pair: having the IDs of both parts server sets the mutual links;
- it's relatively simple when importing certificate for the key already in repo: public key is searched for with the blob extracted from certificate;
- it's not simple when importing key pair for a certificate already existing in repo: parse all certificate and look for corresponding public key ??
- the same if building with the links the certificates chain, when importing certificates in random order...

What do you think about introducing of 'key-identifier' column for CryptographicObjects (here is talking my PKCS#11 background) ? -- with that it should be more simple to manage the links.

viktorTarasov:
updated: use _six_ to resolve _unicode_ in py27 and py34

viktorTarasov:
for Managed CryptoObject is introduced link validator;
PublicKey to link with Certificate is looked for with public key blob obtained from certificate.

PeterHamilton:
@viktorTarasov Managing links for different cryptographic objects can be tricky, depending on the context in which the objects are being created and in which the link(s) would be created or updated.

For public/private keys, it's easy to link the two together if they're created by a call to something like `CreateKeyPair`; just add the link attributes once the keys exist. If the keys are uploaded via `Register`, it is up to the user/client to specify the link attributes, using a template attribute or the `AddAttribute` operation.

For certificates or certificate chains, I would also lean toward requiring the user/client to specify the links directly. There is no mechanism in the KMIP spec for auto-loading or creating a certificate chain in one operation; the closest you can get is to issue a request with a batch of operations, either using `Register` for each certificate in the chain, or constructing the chain directly via a set of `CreateKeyPair` / `Certify` / `Sign` operations.

In general, auto-linking of cryptographic objects (beyond the scope of something like `CreateKeyPair`) is something I want to avoid, for a couple of reasons. (1) It can impose a performance impact on a heavily loaded server when running an exhaustive search for matching key content. (2) It's possible that information could leak about key existence or key ownership, especially if links cross over user object boundaries. This isn't a big problem right now (we don't support explicit user ownership currently) but it's important to keep in mind when we do add that feature.

TL;DR I support adding links, I just err on the side of the user/client explicitly being involved more often than not.

PeterHamilton:
@viktorTarasov Ok, finally had time to look through the patch. First off, thanks for putting so much work into this. I'm impressed with all of the different components you're adding here.

Building off of that, it will be much easier to review and discuss these changes if you break this patch into multiple pull requests, giving how much we're actually dealing with. The logical breakdown I'm thinking of would be something like this:
* a pull request adding the `Link` attribute and tests
* a pull request adding support for `Link` attributes to the attribute factory and tests
* a pull request adding support for `Link` attributes to the Pie object set and tests
* a pull request adding X.509 certificate code to the Cryptography engine and tests
* a pull request adding support for `Link` attributes to the server engine and tests

We may break things down further on some of the bigger chunks (like the server engine) but I think that's a good start. Once these small pull requests are submitted, we can start going through them in detail.

viktorTarasov:
Will be split onto smaller PRs

PR #170, opened 2016-05-16
server-engine: fix crypto-length in Create

viktorTarasov:
replacement for #164

```
In request payload to create symmetric key the
length of cryptographic algorithm is optional.

As Peter suggested, in this case the strongest allowed key size
is used.
```

Difference with the previous PR:
- fixed use of _sizes[-1]_ instead of _sizes[len(sizes) -1]_
- _cryptographic_length_ is added to the crypto/engine result dict
- test for default maximal cryptographic length
- Link init() accepts _enum_ for link type and _str_ for _linked object ID_

Discutable point: introduced proxy to core/factories/attributes in pie/factory.
Motivation for that is to use in pie objects or engines only one pie/factory.

### DIFF
--- a/kmip/pie/factory.py
+++ b/kmip/pie/factory.py
@@ -21,6 +21,9 @@ from kmip.core import secrets
 
 from kmip.pie import objects as pobjects
 
+from kmip.core.factories.attributes import AttributeFactory \
+    as CoreAttributeFactory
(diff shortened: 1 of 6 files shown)


PeterHamilton:
Putting this on ice for now.

PR #171, opened 2016-05-17
Link attribute: add to core attributes

viktorTarasov:
Part of the split PR #167:

_Link_ attribute added to core attributes

PR #172, opened 2016-05-17
Link attribute: factory support

viktorTarasov:
Part of the split PR #167:

Added factory support to create _Link_ core attribute.

Depends on PR #171, where _Link_ core attribute implemented


PeterHamilton:
Putting this on ice for now.

PR #173, opened 2016-05-17
Link attribute: pie SQL type 'CryptographicObjectLink' 

viktorTarasov:
Part of the split PR #167:

Introduces SQL type for pie CryptographicObject link

Depends on PRs #172 and #171, where _Link_ core attribute and it's factory support are implemented


PeterHamilton:
Putting this on ice for now.

PR #174, opened 2016-05-17
Link attribute: pie: 'link' property for crypto objects

viktorTarasov:
Part of the split PR #167:

Implemented method to validate link property for pie cryptographic object,
method to return valid link types for the objects derived from CryptographicObject.

Depends on PRs #173 , #172 and #171, where _Link_ core attribute and factory are implemented, as well as related pie SQL type


PeterHamilton:
Putting this on ice for now.

PR #175, opened 2016-05-18
Link attribute: manage links in server engine

viktorTarasov:
Part of the split PR #167:

In Server Engine creates Link attributes that has or can be created by server:
- mutual links of private and public parts of generated key pair;
- link to PubicKey when registering Certificate;
- support for individual add of Link attribute (for the future implementation of 'AddAttribute' operation).

Depends on PRs #174 , #173 , #172 and #171, where Link core attribute and factory are implemented, related pie SQL type created and link property introduced to PIE crypto objects


viktorTarasov:
last commit rebased:
- remove _'expires_on_commit=False'_  from sqlalchemy session configuration;
- object 'cert' re-instantiated after _session.commit()_

PeterHamilton:
Putting this on ice for now.

PR #169, opened 2016-05-11
'Add Attribute', 'Get Attribute List' and demo 'Import PKCS#12'

viktorTarasov:
Implemented 'add-attribute' and 'get-attribute-list'

Added demo with import PKCS#12.
This part largely uses the LINK attributes, implemented in dependent PR.
This demo creates mutual links between private, public keys and certificate, as well as links for CA certificate chain.
 
It's not final version, but rather the invitation to the rapid review and general consideration.

Q: in demo the pyasn1 and pyOpenSSL is used, is it acceptable? If not, are there any alternatives?

PR #176, opened 2016-05-20
GetAttributeList: dedicated method in pie objects

viktorTarasov:
part of split #169

Introduces support of _GetAttributeList_ operation with the dedicated method of pie _ManagedObject_ and its childs.


viktorTarasov:
rebased to resolve the _pep8_ issues

PeterHamilton:
Putting this on ice for now.

PR #177, opened 2016-05-20
GetAttributeList: process payloads in server engine

viktorTarasov:
part of split #169 

In server engine process the GetAttributeList payloads

depends on #176 where dedicated method is introduced to pie _ManagedObject_ and childs


PeterHamilton:
Putting this on ice for now.

PR #178, opened 2016-06-06
AddAttribute: payload messages

viktorTarasov:
Implementing of "Add Attribute" operation starts with
PR for the core payload messages.

part of work to split #169


PeterHamilton:
Putting this on ice for now.

PR #179, opened 2016-06-06
AddAttribute: payload factories

viktorTarasov:
Second from 'Add Attribute' PR sequence - payload factory support.
Depends on #178 -- 'AddAttribute: payload messages'

part of work to split #169


PeterHamilton:
Putting this on ice for now.

PR #192, opened 2016-09-13
Adding session extraction of client identity from certificates

PeterHamilton:
This change updates the KmipSession, allowing it to extract client identity from the client certificate of a TLS connection. The certificate subject common name is used as the client identity if the certificate has client authentication set in the extended key usage extension.

This change breaks backwards compatibility. If a client certificate does not define a client identity, the session will reject it and shutdown the connection. Any client certificates used to connect with the software server in the past will need to be replaced with certificates that define a suitable client identity.

### DIFF
--- a/kmip/services/server/session.py
+++ b/kmip/services/server/session.py
@@ -18,6 +18,9 @@ import socket
 import struct
 import threading
 
+from cryptography import x509
+from cryptography.hazmat import backends
+
 from kmip.core import enums
 from kmip.core import exceptions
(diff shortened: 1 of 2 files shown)


joel-coffman:
Without having reviewed this code, I'm generally okay with this change.

> This change breaks backwards compatibility.

PyKMIP has yet to release a stable version (1.0.0) so the public API should be considered unstable -- see the [SemVer specification](http://semver.org).

PR #224, opened 2016-12-02
Add optional 'name' argument to create and create_key_pair functions

jseadragon:
The ProxyKmipClient now allows you to optionally provide a name
when performing a `Create` or a `Create Key Pair`. If not specified,
the name is excluded from the request.

 * For `create`, users specify `name`
 * For `create_key_pair`, users specify `private_name` and `public_name`

### DIFF
--- a/kmip/core/factories/attribute_values.py
+++ b/kmip/core/factories/attribute_values.py
@@ -112,10 +112,17 @@ class AttributeValueFactory(object):
 
     def _create_name(self, name):
         if name is not None:
-            name_value = name.name_value
-            name_type = name.name_type
+            if isinstance(name, attributes.Name):
+                return attributes.Name.create(name.name_value, name.name_type)
 
-            return attributes.Name.create(name_value, name_type)
+            elif isinstance(name, str):
+                return attributes.Name.create(
+                            name,
+                            enums.NameType.UNINTERPRETED_TEXT_STRING
+                        )
+            else:
+                raise ValueError('Unrecognized attribute type: '
+                                 '{0}'.format(name))
         else:
             return attributes.Name()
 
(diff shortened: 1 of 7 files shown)


kmip/core/factories/attribute_values.py
```diff
@@ -112,10 +112,17 @@ def create_attribute_value(self, name, value):
...
-
-            return attributes.Name.create(name_value, name_type)
+            if isinstance(name, attributes.Name):
+                return attributes.Name.create(name.name_value, name.name_type)
+
+            elif isinstance(name, str):
+                return attributes.Name.create(
+                            name,
+                            enums.NameType.UNINTERPRETED_TEXT_STRING
+                        )
```
PeterHamilton: Thanks for this. No idea why I didn't have this as default implementation.

kmip/tests/unit/pie/test_client.py
```diff
@@ -261,6 +261,69 @@ def test_create_with_operation_policy_name(self):
...
+        """
+        Test that two identical name attributes match
+        """
+
+        attr_type = enums.AttributeType.NAME
+        attr_name = "foo"
+        attr_a = self.attribute_factory.create_attribute(attr_type, attr_name)
+        attr_b = self.attribute_factory.create_attribute(attr_type, attr_name)
+        self.assertTrue(attr_a == attr_b)
+        self.assertFalse(attr_a != attr_b)
```
PeterHamilton: This test doesn't really belong in `TestProxyKmipClient`. It should go in an attribute factory test suite, but one doesn't exist. I'm fine with leaving this here for now but add a TODO comment indicating it should be moved once the attribute factory test suite is added.

jseadragon: Yeah, I meant to move it but missed it in my original pass through. I'll go ahead and add a attribute factory test suite, not a huge deal. 

Should I do it as a separate commit or squash it?

PeterHamilton: Let's do a separate commit for this one.

jseadragon: Done, accidentally forgot to push the `pep8` fixes i did locally, so it should be up to date now.

PR #238, opened 2016-12-22
Unit test fixes

cfs-pure:
Fix unittests for compatibility between Python 2 and 3.
Fixed a mock for open() in kmip.services.server.server

### DIFF
--- a/kmip/tests/unit/pie/test_client.py
+++ b/kmip/tests/unit/pie/test_client.py
@@ -583,7 +583,7 @@ class TestProxyKmipClient(testtools.TestCase):
             result = client.get_attribute_list(uid)
             client.proxy.get_attribute_list.assert_called_with(uid)
             self.assertIsInstance(result, list)
-            self.assertItemsEqual(attribute_names, result)
+            six.assertCountEqual(self, attribute_names, result)
 
     @mock.patch('kmip.pie.client.KMIPProxy',
                 mock.MagicMock(spec_set=KMIPProxy))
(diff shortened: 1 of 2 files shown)


cfs-pure:
The problems this pull request fixed came up when rebuilding the python-pykmip package from Ubuntu zesty ([0.5.0-4](http://packages.ubuntu.com/source/zesty/python-pykmip)) using the v0.6.0 tarball from PyPI.  The Debian build uses python-coverage and not tox.

PeterHamilton:
I'm fine with the `six.assertCountEqual` and `six.assertRaisesRegex` changes. Can you provide more info on the `mock.patch` change with `create=True`? Was `open` not found when you were running the tests?

cfs-pure:
@PeterHamilton the mock was being applied to the kmip.services.server.server class which does not have an open method.

PeterHamilton:
I think the `open` issue will be better handled if I just switch to using `mock_open`. I'll put out a patch for that. Once that's in, you can pull out the `mock.patch` change and this should be good to go.

PeterHamilton:
Nevermind, I'm already using `mock_open`. The right solution is to scrap the code in the `try` block and just use the right module for Python 3+, `builtins`.

PeterHamilton:
See PR https://github.com/OpenKMIP/PyKMIP/pull/239 for the open mock fix.

PeterHamilton:
PR https://github.com/OpenKMIP/PyKMIP/pull/239 has been merged. Once this PR is updated, I'll merge it as well.

cfs-pure:
Removed my changes to test_server.py because of #239.

PR #246, opened 2017-01-08
Remove support for Python2.6

### LINKED ISSUE #116: Drop support for Python 2.6

PeterHamilton:
This change removes PyKMIP support for Python2.6. The last release of Python2.6 occurred in October 2013, over 3 years ago. It also does not support security features needed to use PyKMIP according to best practices. This change drops tox and TravisCI settings for Python2.6 and updates setup to no longer categorize the library as Python2.6 compliant. The README has been updated accordingly.

Closes #116

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,6 @@ setuptools.setup(
         "Operating System :: POSIX :: Linux",
         "Programming Language :: Python",
         "Programming Language :: Python :: 2",
-        "Programming Language :: Python :: 2.6",
         "Programming Language :: Python :: 2.7",
         "Programming Language :: Python :: 3.3",
         "Programming Language :: Python :: 3.4",
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: python
 matrix:
     include:
-        - python: 2.6
-          env: TOXENV=py26
         - python: 2.7
           env: TOXENV=py27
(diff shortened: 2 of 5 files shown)


setup.py
```diff
@@ -27,10 +27,10 @@
 setuptools.setup(
     name='PyKMIP',
     version=__version__,
-    description='KMIP v1.1 library',
+    description='KMIP library',
```
joel-coffman: This change seems unrelated to Python 2.6.

setup.py
```diff
@@ -27,10 +27,10 @@
...
     keywords='KMIP',
     author='Peter Hamilton',
-    author_email='peter.hamilton@jhuapl.edu',
+    author_email='peter.allen.hamilton@gmail.com',
```
joel-coffman: Ditto

setup.py
PeterHamilton: True, true. Fixed!

setup.py
PeterHamilton: Ditto

PR #253, opened 2017-01-25
Add MAC operation support in cryptography engine

vbnmmnbv:


kmip/services/server/crypto/engine.py
```diff
@@ -148,6 +156,71 @@ def create_asymmetric_key_pair(self, algorithm, length):
...
+        Raises:
+            InvalidField: Raised when the algorithm is unsupported or the
+                length is incompatible with the algorithm.
+            CryptographicFailure: Raised when the key generation process
+                fails.
+
+        Example:
+            >>> engine = CryptographyEngine()
+            >>> mac_data = engine.mac(
+            ...     CryptographicAlgorithm.HMAC-SHA256, b'\x01\x02\x03\x04')
```
PeterHamilton: This example also needs `data`; it's only providing the `algorithm` and `key` right now.

kmip/services/server/crypto/engine.py
```diff
@@ -148,6 +156,71 @@ def create_asymmetric_key_pair(self, algorithm, length):
...
+        """
+
+        mac_data = None
+
+        if algorithm in self._hash_algorithms.keys():
+            self.logger.info(
+                "Generating hash-based Message authentication codes using {0}".
```
PeterHamilton: "Generating hash-based Message authentication codes..." -> "Generating a hash-based message authentication code..."

kmip/services/server/crypto/engine.py
```diff
@@ -148,6 +156,71 @@ def create_asymmetric_key_pair(self, algorithm, length):
...
+                mac_data = h.finalize()
+            except Exception as e:
+                self.logger.exception(e)
+                raise exceptions.CryptographicFailure(
+                    "An error occurred while doing HMAC operation. "
+                    "See the server log for more information."
+                )
+        elif algorithm in self._symmetric_key_algorithms.keys():
+            self.logger.info(
+                "Generating cipher-based Message authentication codes using "
```
PeterHamilton: Ditto.

kmip/services/server/crypto/engine.py
```diff
@@ -148,6 +156,71 @@ def create_asymmetric_key_pair(self, algorithm, length):
...
+            )
+            hash_algorithm = self._hash_algorithms.get(algorithm)
+            try:
+                h = hmac.HMAC(key, hash_algorithm(), backend=default_backend())
+                h.update(data)
+                mac_data = h.finalize()
+            except Exception as e:
+                self.logger.exception(e)
+                raise exceptions.CryptographicFailure(
+                    "An error occurred while doing HMAC operation. "
```
PeterHamilton: "...while doing HMAC..." -> "...while computing an HMAC...

kmip/services/server/crypto/engine.py
```diff
@@ -148,6 +156,71 @@ def create_asymmetric_key_pair(self, algorithm, length):
...
+                "{0}".format(algorithm.name)
+            )
+            cipher_algorithm = self._symmetric_key_algorithms.get(algorithm)
+            try:
+                c = cmac.CMAC(cipher_algorithm(key), backend=default_backend())
+                c.update(data)
+                mac_data = c.finalize()
+            except Exception as e:
+                raise exceptions.CryptographicFailure(
+                    "An error occurred while doing CMAC operation. "
```
PeterHamilton: Ditto.

kmip/services/server/crypto/engine.py
```diff
@@ -148,6 +156,71 @@ def create_asymmetric_key_pair(self, algorithm, length):
...
+                    "See the server log for more information."
+                )
+        else:
+            raise exceptions.InvalidField(
+                "The cryptographic algorithm ({0}) is not a supported "
+                "for MAC operation.".format(algorithm)
```
PeterHamilton: "...for MAC operation." -> "...for a MAC operation."

kmip/services/server/crypto/engine.py
```diff
@@ -148,6 +156,71 @@ def create_asymmetric_key_pair(self, algorithm, length):
...
+                    "See the server log for more information."
+                )
+        elif algorithm in self._symmetric_key_algorithms.keys():
+            self.logger.info(
+                "Generating cipher-based Message authentication codes using "
+                "{0}".format(algorithm.name)
+            )
+            cipher_algorithm = self._symmetric_key_algorithms.get(algorithm)
+            try:
+                c = cmac.CMAC(cipher_algorithm(key), backend=default_backend())
```
PeterHamilton: Nice job here including CMAC in addition to HMAC. The KMIP specification doesn't indicate what specific type of MAC operation to support, so anything possible in the list of CryptographicAlgorithms is valid.

We should probably add a comment here indicating that the ARC4 and IDEA algorithms aren't compatible with CMAC; CMAC requires block ciphers and those two aren't block ciphers.

kmip/tests/unit/services/server/crypto/test_engine.py
```diff
@@ -146,3 +146,123 @@ def test_create_asymmetric_key_with_invalid_length(self):
...
+
+    def test_mac(self):
+        """
+        Test that MAC operation can be done with valid arguments.
+        """
+        key1 = (b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
+                b'\x00\x00\x00\x00')
+        key2 = (b'\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
+                b'\x00\x00\x00\x00')
+        key3 = key1
```
PeterHamilton: Go ahead and actually duplicate the key1 bytes here for key3. That way we explicitly show that the same key bytes, in completely separate objects, yield the same MAC for the same input data.

kmip/tests/unit/services/server/crypto/test_engine.py
```diff
@@ -146,3 +146,123 @@ def test_create_asymmetric_key_with_invalid_length(self):
...
+
+    def test_mac_with_cryptographic_failure(self):
+        pass
+        """
+        Test that an CryptographicFailure error is raised when the MAC
+        process fails.
+        """
+
+        # Create dummy algorithm that always fails on instantiation.
+        class DummySymmetricKeyAlgorithm(object):
```
PeterHamilton: You should be able to use ARC4 or IDEA here instead of needing this dummy. Those should automatically fail with CMAC.

kmip/services/server/crypto/engine.py
vbnmmnbv: Right~

kmip/services/server/crypto/engine.py
vbnmmnbv: Right. I added a comment. Please check whether it's fine

vbnmmnbv:
Hi Peter,

I tried to address all your comments. Then I did an amend and force push.

It seems that all your previous comments are gone by this way? Not sure whether I did the right thing or not.

PeterHamilton:
Not a problem, I can still see my comments as part of the review on the prior version.

PR #254, opened 2017-01-25
Add MAC operation support at client side

vbnmmnbv:
Hi, 

I make it as one commit.
Since I mostly care about the pie api:
(1). I didn't add any tests in test_kmip_client.py
(2). I didn't make any example in demos/units/
Please let me know if there is any concern.

Once I got this merged, I will add the server engine part.


vbnmmnbv:
Does the coverage testing result mean my test is not thorough enough? Is it a blocker?

PeterHamilton:
Yes, it is a blocker. Go ahead and add the tests that are needed. I'm fine with adding demos later on.

vbnmmnbv:
Working on it.....

vbnmmnbv:
Finally passed the test coverage ...... 
Only missing test for mac() in kmip_client.py, where many functions were not tested currently~

kmip/tests/unit/core/attributes/test_attributes.py
```diff
@@ -498,25 +501,33 @@ def setUp(self):
...
 
         self.cp_none = self.factory.create_attribute_value(
             AttributeType.CRYPTOGRAPHIC_PARAMETERS, {})
 
         # Symmetric key object with Cryptographic Parameters
-        # Byte stream edited to add Key Role Type parameter
+        # Byte stream edited to add:
+        #      Key Role Type parameter
+        #      Digital Signature Algorithm parameter
+        #      Cryptographic Algorithm parameter
```
PeterHamilton: Did you get the bytes from a specific test case document?

kmip/tests/unit/core/messages/test_messages.py
```diff
@@ -1775,6 +1963,175 @@ def test_locate_response_write(self):
...
+        result = self.stream.read()
+        len_exp = len(self.mac)
+        len_rcv = len(result)
+        self.assertEqual(len_exp, len_rcv,
+                         self.msg.format('get response message', 'write',
+                                         len_exp, len_rcv))
+
+        msg = "Bad response message write: encoding mismatch"
+        self.assertEqual(self.mac, result, msg)
+
```
PeterHamilton: I'm fine with keeping these additions to `test_messages.py`. However we do need a dedicated `payloads/test_mac.py` just for the MAC payloads. See [GetAttributesList](https://github.com/OpenKMIP/PyKMIP/blob/master/kmip/tests/unit/core/messages/payloads/test_get_attribute_list.py) and [GetAttributes](https://github.com/OpenKMIP/PyKMIP/blob/master/kmip/tests/unit/core/messages/payloads/test_get_attributes.py) for examples.

kmip/core/attributes.py
```diff
@@ -354,6 +379,21 @@ def __validate(self):
                 msg += "; expected {0}, received {1}".format(
                     self.KeyRoleType, self.key_role_type)
                 raise TypeError(msg)
+        if self.digital_signature_algorithm is not None:
+            if not isinstance(self.digital_signature_algorithm,
+                              self.DigitalSignatureAlgorithm):
```
vbnmmnbv: I do not quite understand.  DigitalSignatureAlgorithm is a class defined in CryptographicParameters just like KeyRoleType and etc.... What's the problem for using "self.DigitalSignatureAlgorithm"?

kmip/core/attributes.py
```diff
@@ -354,6 +379,21 @@ def __validate(self):
...
+                msg = "Invalid digital signature algorithm"
+                msg += "; expected {0}, received {1}".format(
+                    self.DigitalSignatureAlgorithm,
```
vbnmmnbv: Ditto. Please give me more details

kmip/tests/unit/core/attributes/test_attributes.py
vbnmmnbv: I didn't find any MAC related test data here:
http://docs.oasis-open.org/kmip/testcases/v1.2/kmip-testcases-v1.2.html
Do you know where I should look?
So unfortunately, the data is generated by my code logic and then used to test my code itself.....

kmip/core/attributes.py
PeterHamilton: You are correct in observing that this is done in other parts of the library. You are also correct that both versions are effectively the same; however, there is a difference. `self` refers directly to the instance of `CryptographicParameters` (which may vary from a vanilla instance depending on what's done with it by the instantiating code), while specifying `CryptographicParameters.` is a direct reference to the class definition (which is always what we want here). This actually ties into Python metaclass mechanics and how attributes are handled and accessed but that's a deep rabbit hole. To quote the Zen of Python, "Explicit is better than implicit."

I used to not really care about explicit vs implicit references. However, this ended up causing a nasty self reflection bug way back in an earlier version of the library (two years ago or so). Now, I try to avoid overusing `self`. One of my (many) TODOs is to go through and fix these references to explicitly use class names; I've just never gotten around to it. I also plan on pulling out these internal class definitions (e.g., `BlockCipherMode`, `PaddingMethod`, etc) and placing them in the outer module scope. Again, one of my many TODOs.

kmip/tests/unit/core/attributes/test_attributes.py
PeterHamilton: Hmmm, that's what I was worried about. I'll try to find examples in the other test case documents.

If I can't find any I'll go through the bytes manually and confirm they're valid.

kmip/core/messages/payloads/mac.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def __init__(self,
+                 unique_identifier=None,
+                 cryptographic_parameters=None,
+                 data=None):
+        super(MACRequestPayload, self).__init__(
+            tag=enums.Tags.REQUEST_PAYLOAD)
+        self.unique_identifier = unique_identifier
+        self.cryptographic_parameters = cryptographic_parameters
+        self.data = data
```
vbnmmnbv: Hi, I have a concern for unique_identifier. It seems that for all the other payloads, the input is type attribute.UniqueIdentifier, while for the getattribute and getattributelist payloads, the input is of type string. Do you think it's not consistent?

vbnmmnbv: In my mind, probably it is more clear to keep the uuid input to all payload class as type attribute.UniqueIdentifier?

Will also apply to Data and MACData~

kmip/core/attributes.py
vbnmmnbv: Sure. I will modify my part and leave yours intact~

kmip/core/messages/payloads/mac.py
vbnmmnbv: Actually, can I have an argument with you about the necessity of using property here? Maybe for getattribut is necessary. But except type checking(e.g., whether cryptographic_parameters is really type CryptographicParameters object or not), I can not see any really meaningful check here for MAC payload. The type correctness is also mostly guaranteed by client.py/kmip_client.py.  So do you still prefer to use property here?

kmip/tests/unit/core/attributes/test_attributes.py
PeterHamilton: Alright, the encoding looks correct. I'm ok with this.

kmip/core/messages/payloads/mac.py
PeterHamilton: It is inconsistent, but there's little risk here. The clients can either handle these payloads differently or multiple types can be handled in the property code itself, whichever we prefer. The benefit here to using properties is that it simplifies the use of the payloads long term, merging how a caller sees the `unique_identifier` as a string with its internal representation as a serializable structure.

Moreover, now there's no risk of missing or forgetting a `validate` call, since all attribute access is implicitly type-checked. For example, if a caller instantiated the payload, and then modified the attributes directly, it is very likely they would not call `validate` once they were done to catch mistakes. Now, they know instantly if they made a mistake, any time they access one of these attributes.

As for debug-only builds, we really don't have any structure in place to do that in the library. Even if we did, we would still want to verify attribute types since user/client code could easily pass in bad value types and trigger lower-level errors (or worse, generate a "valid" serializable request that actually doesn't do what the user/client thinks it should do). If performance is really a concern for someone's application, they shouldn't be using Python.

vbnmmnbv: Fair enough. I am currently working on it. But want to bring up something to ensure my understanding  is correct. In GetAttributeListResponsePayload,  in read/write functions, you did not raise exception when required fields (i.e., uuid and attribute name according to spec) are missing. This is different from what you did in getattribute. That's something you missed, right?

PeterHamilton: Good eye. The answer is "sort of." If you look through the library you'll find lots of places where parse errors can occur, raising exceptions. This is pretty standard in OOP. The problem is that this makes the code harder to work with. For example, the user/client may provide an extra unneeded attribute value, or instead omit one, when making a client call. This mistake may not trigger an exception until way down the processing stack when reading or writing an entire request/response structure. This is frustrating to the user because they get an error related to library internals that they don't understand, and its frustrating for developers because the source of the error is far away from where the error is actually detected, making it hard to find and fix.

There are a couple of solutions to this. The obvious one is to just put in better error handling. The problem is this would pretty much have to happen everywhere and it would even further bloat and complicate the code. An alternative (which is what I explored for a time) was to rethink what is actually a parse error. The idea here is that if a request/response message can be parsed successfully (i.e., the fields present are formatted properly and there are no partial fields that can't get read), let it get parsed. If the payload is missing certain fields, let the user/client know by a more organic, less intrusive mechanism than just erroring out. I started playing around with this with the `GetAttributeListResponsePayload`, to see if I liked how it felt/worked and if it caused any problems with active deployments. I don't have enough "data" yet to determine if it's a viable way forward, so I'm not sure if I'll stick with it.

For now, with the MAC payloads, I'm fine with you putting in those read/write exceptions for missing required fields. If need be, I'll patch in the right exception(s) needed for `GetAttributeList` and make everything consistent again.

vbnmmnbv: Thanks for  the explanation. Good to learn~

vbnmmnbv:
My coverage failed because of the mac function in kmip_client.py. I noticed most functions here are not tested. So I just want to confirm with you about what's the right way to test these functions.
Do you think it's reasonable to  mock send/receive functions and check whether it constructs the right message to send and can parse the given response as expected?

PeterHamilton:
I think that's reasonable. Some of the operations have some testing support, like `Get`, `CreateKeyPair`, and `DiscoverVersions`, so you can use those as inspiration.

The long term goal is to merge the two clients and drop a lot of the old code that's in the `KMIPProxy`. Any tests required for new patch coverage are sufficient for now.

vbnmmnbv:
passed!

PR #255, opened 2017-01-30
Add necessary objects/attributes and payloads for MAC operation

vbnmmnbv:


vbnmmnbv:
So I will split the attributes/payload from client codes

PeterHamilton:
Nice, I'll go through it real quick. Be sure to rebase #254 to reflect this.

PR #256, opened 2017-02-01
Add MAC operation support for server

vbnmmnbv:


kmip/services/server/engine.py
```diff
@@ -1556,3 +1561,61 @@ def _process_discover_versions(self, payload):
...
+
+        unique_identifier = self._id_placeholder
+        if payload.unique_identifier:
+            unique_identifier = payload.unique_identifier.value
+
+        object_type = self._get_object_type(unique_identifier)
+
+        managed_object = self._data_session.query(object_type).filter(
+            object_type.unique_identifier == unique_identifier
+        ).one()
```
PeterHamilton: Replace 1573-1577 with a call to `_get_object_with_access_controls`; see other operations, like `_process_get` or `_process_destroy` for examples.

vbnmmnbv: So that's the new stuff~ But I met an issue.  It seems that the operation policy(3.18) doesn't control operations like MAC which should be restricted by crypto masks(as we discussed before). That's why in my codes I didn't check policy in mac operation. In the new function, _is_allowed_by_operation_policy is called anyway. So do you think I should add MAC into policy.py or keep my current codes?

PeterHamilton: In this case I think it's appropriate to pass `GET` as the operation to `_get_object_with_access_controls` instead of `MAC`. That way we ensure that only a user with read access to the secret can use it to compute a MAC value.

vbnmmnbv:
Done. The whole operation works before when I did my first commit. Hopefully I didn't break anything.

Issue #258, opened 2017-02-06
about locate operaton


vbnmmnbv:
Hi Peter,

Can I ask you two questions?

1. Do you have any ETA for the next release?

2. Currently I need an operation to "list" all the existing keys on the server. Should be "locate". Seems that currently it's not supported by server. 
I checked the SPEC, maximum items/storage status mask/object group member are all non-trivial features to support, I think. 
So do you think currently it makes sense to implement a simple locate operation which will just return uuids of all the existing managed objects of the server without any filtering? Later we can add the filtering piece by piece.

PeterHamilton:
1. I usually cut a new release every 4-5 months or so, depending on how much work has been done with the library in that time. 0.7.0 will probably come out sometime in April.

2. Yes, a full-featured version of `locate` is challenging to support. I'm fine with a simple, bare bones version that only lists the UUIDs of the objects belonging to a specific user. That's a reasonable first cut on it.

PR #260, opened 2017-02-09
Add minimum Locate operation support for server.

vbnmmnbv:
Currently it will return all the existing managed objects with the restriction of operation
policy only. No other filterings.

### DIFF
--- a/kmip/services/server/engine.py
+++ b/kmip/services/server/engine.py
@@ -49,6 +49,7 @@ from kmip.core.messages.payloads import get_attribute_list
 from kmip.core.messages.payloads import query
 from kmip.core.messages.payloads import register
 from kmip.core.messages.payloads import mac
+from kmip.core.messages.payloads import locate
 
 from kmip.core import misc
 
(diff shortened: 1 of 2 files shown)


kmip/tests/unit/services/server/test_engine.py
```diff
@@ -3461,6 +3462,91 @@ def test_request_omitting_secret(self):
...
+           2
+        )
+        self.assertEqual(
+            id_a,
+            response_payload.unique_identifiers[0].value
+        )
+        self.assertEqual(
+            id_b,
+            response_payload.unique_identifiers[1].value
+        )
```
PeterHamilton: These last two asserts should be changed to `assertIn` checks. There's no guarantee that SQLAlchemy will return the objects in the order they were added.

PeterHamilton:
Oh, and add `Closes #258` to the bottom of your commit message (see #246 as an example). This will close the corresponding issue you filed earlier.

kmip/services/server/engine.py
```diff
@@ -1299,6 +1324,27 @@ def _process_register(self, payload):
 
         return response_payload
 
+    @_kmip_version_supported('1.2')
```
vbnmmnbv: Th reason why I did this is that I found in 1.0, there is no "object group member" in the request. So do we want to change it to be 1.1 instead of 1.0?

kmip/tests/unit/services/server/test_engine.py
vbnmmnbv: Got it. I will change it.

vbnmmnbv:
For the "Closes #258", should I add it in this commit or add it to the commit(all Locate pie support) currently I am working on?

kmip/services/server/engine.py
PeterHamilton: No, `1.0` is fine. It's up to the implementation for `Locate` to handle the different possible payload versions.

PeterHamilton:
Add it to whichever commit addresses and "finishes" the issue.

vbnmmnbv:
Please bear with my ability of understanding a little bit~ So in this case(I have a server change and then pie change), I should add it to the pie change only?

PeterHamilton:
Ah, in that case add it to this commit, since the issue was about adding `Locate` support to the server.

Issue #261, opened 2017-02-10
client.py argument checking


vbnmmnbv:
Hi Peter,

When I am working on adding the pie support for locate operation, I do notice that there are some issues about the arguments checking in client.py. It's too strict, I think. For example, for get() operation,  from the spec, uid can be ignored  while it must be a string in the codes. This means that using the pie interface, I can not issue an get request without any uid? 

Should we refactor this some time? I think your vision is to put all the argument validation inside the payload module in the future?

PeterHamilton:
Good point, `Get` and `Destroy` should allow `None`. The underlying `KMIPProxy` client already works this way. I've pushed a patch updating these.

PeterHamilton:
I need to rebase #262 before this is closed. I'll do that tomorrow.

PR #267, opened 2017-03-30
Add Activate operation support for pie client

vbnmmnbv:
Hi Peter, for the last pull request about pie activate support (you have already merged), I realized that I made the same mistake as 
https://github.com/OpenKMIP/PyKMIP/issues/261

So I fixed it.  Can you add this amend?

BTW, did the fix you made for 261 merged?


PeterHamilton:
Don't worry about it. Just create a new patch that just adds in the `None` check. No need to back out the prior patch and merge this one.

PR #268, opened 2017-03-30
fix activate uid none check

vbnmmnbv:


vbnmmnbv:
So I only add the fix for activate operation.
For get/destroy, I have seen you already have fix that has not been merged yet~

PR #271, opened 2017-03-31
Add initial Revoke operation support for server.

vbnmmnbv:
See the TODO in _process_revoke for more info.

### DIFF
--- a/kmip/services/server/engine.py
+++ b/kmip/services/server/engine.py
@@ -39,6 +39,7 @@ from kmip.core.messages import contents
 from kmip.core.messages import messages
 
 from kmip.core.messages.payloads import activate
+from kmip.core.messages.payloads import revoke
 from kmip.core.messages.payloads import create
 from kmip.core.messages.payloads import create_key_pair
 from kmip.core.messages.payloads import destroy
@@ -971,6 +972,8 @@ class KmipEngine(object):
             return self._process_get_attribute_list(payload)
         elif operation == enums.Operation.ACTIVATE:
             return self._process_activate(payload)
+        elif operation == enums.Operation.REVOKE:
+            return self._process_revoke(payload)
         elif operation == enums.Operation.DESTROY:
             return self._process_destroy(payload)
         elif operation == enums.Operation.QUERY:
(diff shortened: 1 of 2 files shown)


kmip/services/server/engine.py
```diff
@@ -1501,6 +1504,55 @@ def _process_activate(self, payload):
...
+                        [x.capitalize() for x in object_type.name.split('_')]
+                    )
+                )
+            )
+
+        # TODO: (1) need to set Compromise Date attribute or Deactivation Date
+        # attribute, (2) The revoke (KMIP1.2 4.20)  doesn't say anything about
+        # the state transition permission, but the state (KMIP1.2 3.22) diagram
+        # does show that not all states can transit to "Deactivated" or
+        # "Compromised". So what's the right way?
```
PeterHamilton: Yeah, the info for this is spread around Sections 3.22 and 4.20. This is how the state transition should go.

If the reason code is `KEY_COMPROMISE`:
`PRE_ACTIVE` -> `COMPROMISED`
`ACTIVE` -> `COMPROMISED`
`DEACTIVATED` -> `COMPROMISED`
`DESTROYED` -> `DESTROYED_COMPROMISED`

If the reason code is *not* `KEY_COMPROMISE`:
`ACTIVE` -> `DEACTIVATED`

If someone tries to revoke an object that is not in the `ACTIVE` state, for a reason that is *not* a compromise, that should be an error (`IllegalOperation` seems fine).

kmip/services/server/engine.py
```diff
@@ -1501,6 +1504,55 @@ def _process_activate(self, payload):
...
+        if revocation_code.value is enums.RevocationReasonCode.KEY_COMPROMISE:
+            managed_object.state = enums.State.COMPROMISED
```
PeterHamilton: Let's update this to match the cases in my comment above. Error handling included if the current state can't support revocation.

kmip/services/server/engine.py
```diff
@@ -1501,6 +1504,55 @@ def _process_activate(self, payload):
...
+        else:
+            managed_object.state = enums.State.DEACTIVATED
```
PeterHamilton: Ditto.

kmip/tests/unit/services/server/test_engine.py
```diff
@@ -4361,6 +4363,199 @@ def test_activate_not_allowed_by_policy(self):
...
+            response_payload.unique_identifier.value
+        )
+
+        symmetric_key = e._data_session.query(
+            pie_objects.SymmetricKey
+        ).filter(
+            pie_objects.ManagedObject.unique_identifier == object_id
+        ).one()
+
+        self.assertEqual(enums.State.DEACTIVATED, symmetric_key.state)
```
PeterHamilton: The whole test looks good. Just update it to test for those specific state cases. Feel free to break this into smaller test functions too if you want.

kmip/tests/unit/services/server/test_engine.py
```diff
@@ -4361,6 +4363,199 @@ def test_activate_not_allowed_by_policy(self):
...
+
+    def test_revoke_on_static_object(self):
+        """
+        Test that the right error is generated when an reovke request is
```
PeterHamilton: reovke -> revoke

kmip/services/server/engine.py
vbnmmnbv: So basically every state can go into "Compromised" state while only "active" can go into "deactivated" state. Right?

kmip/services/server/engine.py
```diff
@@ -1501,6 +1504,58 @@ def _process_activate(self, payload):
...
+                    ''.join(
+                        [x.capitalize() for x in object_type.name.split('_')]
+                    )
+                )
+            )
+
+        # TODO: need to set Compromise Date attribute or Deactivation Date
+        # attribute
+        if revocation_code.value is enums.RevocationReasonCode.KEY_COMPROMISE:
+            managed_object.state = enums.State.COMPROMISED
```
PeterHamilton: There's one more case that needs to be handled here. If the object state is `DESTROYED` here, its new state should be `DESTROYED_COMPROMISED`. The server currently auto deletes destroyed objects so it's not an immediate issue, but this could change in the future and I'd like to cover that case if it ever comes up.

kmip/tests/unit/services/server/test_engine.py
```diff
@@ -4361,6 +4363,248 @@ def test_activate_not_allowed_by_policy(self):
...
+
+        object_id = str(managed_object.unique_identifier)
+
+        reason_unspecified = objects.RevocationReason(
+            code=enums.RevocationReasonCode.UNSPECIFIED)
+        date = primitives.DateTime(
+            tag=enums.Tags.COMPROMISE_OCCURRENCE_DATE, value=6)
+
+        # Test that reason UNSPECIFIED will put object into state
+        # DEACTIVATED
```
PeterHamilton: Update this comment to reflect the following exception.

kmip/tests/unit/services/server/test_engine.py
```diff
@@ -4361,6 +4363,248 @@ def test_activate_not_allowed_by_policy(self):
...
+
+        args = (payload, )
+        regex = "The object is not active and cannot be revoked with " \
+                "reason other than KEY_COMPROMISE"
+        self.assertRaisesRegexp(
+            exceptions.IllegalOperation,
+            regex,
+            e._process_revoke,
+            *args
+        )
```
PeterHamilton: This test looks good, just update it to handle the `DESTROYED` -> `DESTROYED_COMPROMISED` case.

kmip/tests/unit/services/server/test_engine.py
vbnmmnbv: Yeah, this is a mistake.

vbnmmnbv: So the test goal of this function is already in the function comment at top. I will just delete this comment.

PR #273, opened 2017-04-01
Add state/crypto usage mask checks for MAC operation

vbnmmnbv:


kmip/services/server/engine.py
```diff
@@ -1634,7 +1634,7 @@ def _process_mac(self, payload):
               managed_object.cryptographic_algorithm):
             algorithm = managed_object.cryptographic_algorithm
         else:
-            raise exceptions.InvalidField(
+            raise exceptions.PermissionDenied(
```
PeterHamilton: This looks like it matches what KMIP 1.3 says is the right error for this case, so I'm fine with this.

kmip/services/server/engine.py
```diff
@@ -1652,10 +1652,22 @@ def _process_mac(self, payload):
...
+        if managed_object.state != enums.State.ACTIVE:
+            raise exceptions.PermissionDenied(
+                "Object is not in a state that can be used for MACing."
+            )
+
+        if enums.CryptographicUsageMask.MAC_GENERATE not in \
+           managed_object.cryptographic_usage_masks:
+                raise exceptions.PermissionDenied(
+                    "MAC GERERATE crypto usage mask must be set "
+                    "for the MAC operation"
```
PeterHamilton: Let's change the error message to: "MAC Generate must be set in the object's cryptographic usage mask"

PR #274, opened 2017-04-03
Add state check for Destroy operation.

vbnmmnbv:
Only can be destroyed if in "PRE_ACTIVE" or "DEACTIVATED" state

### DIFF
--- a/kmip/services/server/engine.py
+++ b/kmip/services/server/engine.py
@@ -1509,19 +1509,27 @@ class KmipEngine(object):
         else:
             unique_identifier = self._id_placeholder
 
-        self._get_object_with_access_controls(
+        managed_object = self._get_object_with_access_controls(
             unique_identifier,
             enums.Operation.DESTROY
         )
 
-        # TODO (peterhamilton) Process attributes to see if destroy possible
-        # 1. Check object state. If invalid, error out.
-        # 2. Check object deactivation date. If invalid, error out.
+        # TODO If in "ACTIVE" state, we need to check its "Deactivation date"
+        # to see whether it can be destroyed or not
+        if hasattr(managed_object, 'state'):
+            if managed_object.state == enums.State.ACTIVE:
+                raise exceptions.PermissionDenied(
+                    "Object is active and cannot be destroyed."
+                )
+
+        # 'OpaqueObject' object has no attribute 'state'
+        if hasattr(managed_object, 'state') and \
+           managed_object.state == enums.State.COMPROMISED:
+            managed_object.state = enums.State.DESTROYED_COMPROMISED
 
         self._logger.info(
             "Destroying an object with ID: {0}".format(unique_identifier)
         )
-
         self._data_session.query(objects.ManagedObject).filter(
             objects.ManagedObject.unique_identifier == unique_identifier
         ).delete()
--- a/kmip/tests/unit/services/server/test_engine.py
+++ b/kmip/tests/unit/services/server/test_engine.py
@@ -4373,14 +4373,21 @@ class TestKmipEngine(testtools.TestCase):
 
         obj_a = pie_objects.OpaqueObject(b'', enums.OpaqueDataType.NONE)
         obj_b = pie_objects.OpaqueObject(b'', enums.OpaqueDataType.NONE)
+        key = (b'\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
+               b'\x00\x00\x00\x00\x00')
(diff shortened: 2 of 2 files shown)


kmip/services/server/engine.py
```diff
@@ -1509,14 +1509,20 @@ def _process_destroy(self, payload):
...
-        # 2. Check object deactivation date. If invalid, error out.
+        # TODO If in "ACTIVE" state, we need to check its "Deactivation date"
+        # to see whether it can be destroyed or not
+        if hasattr(managed_object, 'state'):
+            if managed_object.state != enums.State.PRE_ACTIVE and \
+               managed_object.state != enums.State.DEACTIVATED:
+                raise exceptions.PermissionDenied(
+                    "Object is not pre-active or deactivated "
+                    "and cannot be destroyed."
+                )
```
PeterHamilton: Let's also handle the case where if the state is `COMPROMISED` we move it to `DESTROYED COMPROMISED`.

kmip/tests/unit/services/server/test_engine.py
```diff
@@ -4468,6 +4468,45 @@ def test_destroy_not_allowed_by_policy(self):
...
+        regex = "Object is not pre-active or deactivated and cannot be " \
+                "destroyed."
+        six.assertRaisesRegex(
+            self,
+            exceptions.PermissionDenied,
+            regex,
+            e._process_destroy,
+            *args
+        )
+
```
PeterHamilton: Just update this to handle the the `COMPROMISED` state. Might make more sense to add it as a separate test function.

kmip/services/server/engine.py
vbnmmnbv: Please comment on the log message I use here.

PeterHamilton:
Once this gets rebased it should be good to go.

vbnmmnbv:
Done~

kmip/services/server/engine.py
```diff
@@ -1509,22 +1509,37 @@ def _process_destroy(self, payload):
...
-
-        self._logger.info(
-            "Destroying an object with ID: {0}".format(unique_identifier)
-        )
+        # TODO If in "ACTIVE" state, we need to check its "Deactivation date"
+        # to see whether it can be destroyed or not
+        if hasattr(managed_object, 'state'):
+            if managed_object.state != enums.State.PRE_ACTIVE and \
+               managed_object.state != enums.State.DEACTIVATED and \
+               managed_object.state != enums.State.COMPROMISED:
```
PeterHamilton: Let's just change this compound if-statement to a `managed_object.state in [...]` check. It's a little cleaner.

PeterHamilton: It might also just be easier to do an `ACTIVE` check here instead, since that's the only state that can't be directly destroyed.

kmip/services/server/engine.py
```diff
@@ -1509,22 +1509,37 @@ def _process_destroy(self, payload):
...
+                "Destroyed Compromised state)".format(unique_identifier)
+            )
+            managed_object.state = enums.State.DESTROYED_COMPROMISED
+        else:
+            self._logger.info(
+                "Destroying an object with ID: {0}".format(unique_identifier)
+            )
+            self._data_session.query(objects.ManagedObject).filter(
+                objects.ManagedObject.unique_identifier == unique_identifier
+            ).delete()
```
PeterHamilton: Let's condense this a little bit. Let's remove the first log statement, so the `if` just sets `DESTROYED_COMPROMISED`. Then we can drop the `else:` line and keep the rest, letting the log and delete lines happen regardless of state. That way, `DESTROYED_COMPROMISED` objects still get deleted and the logs are consistent. This also makes it easier if we want to add special logic around when we actually delete objects down the road.

kmip/tests/unit/services/server/test_engine.py
```diff
@@ -4435,6 +4435,55 @@ def test_destroy(self):
...
+        symmetric_key = e._data_session.query(
+            pie_objects.SymmetricKey
+        ).filter(
+            pie_objects.ManagedObject.unique_identifier == object_id
+        ).one()
+
+        self.assertEqual(enums.State.DESTROYED_COMPROMISED,
+                         symmetric_key.state)
+
```
PeterHamilton: I see how this ties in to how delete handled `COMPROMISED` objects above. I think it's ok to collapse this down into just a delete call on a `COMPROMISED` object. That code branch in delete will still get covered. We can test this explicitly later when we add more granularity on when to delete objects.

Issue #265, opened 2017-03-29
create/create_key_pair operation in pie


vbnmmnbv:
Hi Peter,

Maybe I have some misunderstanding of your philosophy of pie, but I think for create/create_key_pair, it's not vey reasonable to only have "name"  and "policy" as the attributes inputs.

 In real scenario, creating a key with other attributes probably is very common. For example, for MAC, I need to set the crypto usage mask for the key in the create() operation.  Or maybe I want to tag a "activate date" with it.  Or a "custom attribute" with it. Of course I can add the attributes later but that's not very neat and also make the operation not atomic.

So do you think it's more reasonable to give them a list of attributes as the input instead of just "name" and "policy". 

I know this will make the user to build all the attributes. But I think that's necessary step to make the pie interface really useful in real scenario.

vbnmmnbv:
If you agree, I am happy to work on it :)

PeterHamilton:
You're not wrong. The original intent for `pie` was to make a simple, obvious interface for KMIP operations. The original `KMIPProxy` requires you to prebuild your attribute objects manually, which is messy and not well documented. It also exposes way too much of the internal library guts, which can change. The plan was always to add on arguments to the `ProxyKmipClient` methods as needed, so adding something like `cryptographic_usage_mask` is fine.

I've toyed with just adding a `**kwargs` field to each client method and calling it a day, but I prefer having explicit arguments. They're self documenting and it just makes things clearer for developers. Of course, with something like `create_key_pair` this could get really messy, since you theoretically could have tons of unique attributes for both public and private keys.

What/how many specific arguments do you need and for which operations?

vbnmmnbv:
Currently in my project, I probably want to add crypto usage mask and "custom attributes" to the create() operation. 

So as a tradeoff, do you think it's a  reasonable design to add an additional argument for create():

`  def create(self, algorithm, length, operation_policy_name=None, name=None, other_attributes=None)`

'other_attributes' is a list and need to be built by the user if he wants to add additional attributes.

Similar to create_key_pair:

```
    def create_key_pair(self,
                        algorithm,
                        length,
                        operation_policy_name=None,
                        public_name=None,
                        private_name=None,
                        other_common_attributes=None,
                        other_public_attributes=None
                        other_private_attributes=None
):
```

PeterHamilton:
[...]

PR #276, opened 2017-04-04
Add crypto_usage_masks argument to pie Create operation

vbnmmnbv:
related to #265

### DIFF
--- a/kmip/pie/client.py
+++ b/kmip/pie/client.py
@@ -138,7 +138,8 @@ class ProxyKmipClient(api.KmipClient):
                 self.logger.exception("could not close client connection", e)
                 raise e
 
-    def create(self, algorithm, length, operation_policy_name=None, name=None):
+    def create(self, algorithm, length, operation_policy_name=None, name=None,
+               cryptographic_usage_mask=None):
         """
         Create a symmetric key on a KMIP appliance.
 
@@ -149,6 +150,9 @@ class ProxyKmipClient(api.KmipClient):
             operation_policy_name (string): The name of the operation policy
                 to use for the new symmetric key. Optional, defaults to None
             name (string): The name to give the key. Optional, defaults to None
+            cryptographic_usage_mask (list): list of enumerations of crypto
+                usage mask passing to the symmetric key. Optional, defaults to
+                None
 
         Returns:
             string: The uid of the newly created symmetric key.
@@ -164,6 +168,13 @@ class ProxyKmipClient(api.KmipClient):
                 "algorithm must be a CryptographicAlgorithm enumeration")
         elif not isinstance(length, six.integer_types) or length <= 0:
             raise TypeError("length must be a positive integer")
+        if cryptographic_usage_mask is not None:
+            if not isinstance(cryptographic_usage_mask, list) or \
+                all(isinstance(item, enums.CryptographicUsageMask)
+                    for item in cryptographic_usage_mask) is False:
+                raise TypeError(
+                    "cryptographic_usage_mask must be a list of "
+                    "CryptographicUsageMask enumerations")
 
         # Verify that operations can be given at this time
         if not self._is_open:
@@ -173,7 +184,8 @@ class ProxyKmipClient(api.KmipClient):
         common_attributes = self._build_common_attributes(
             operation_policy_name
         )
-        key_attributes = self._build_key_attributes(algorithm, length)
+        key_attributes = self._build_key_attributes(
+                            algorithm, length, cryptographic_usage_mask)
         key_attributes.extend(common_attributes)
 
         if name:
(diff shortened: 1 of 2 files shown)


kmip/pie/client.py
```diff
@@ -138,7 +138,8 @@ def close(self):
                 self.logger.exception("could not close client connection", e)
                 raise e
 
-    def create(self, algorithm, length, operation_policy_name=None, name=None):
+    def create(self, algorithm, length, operation_policy_name=None, name=None,
+               crypto_usage_masks=list()):
```
PeterHamilton: I'm fine using the full name:
`crypto_usage_masks` -> `cryptographic_usage_mask`

kmip/pie/client.py
```diff
@@ -149,6 +150,8 @@ def create(self, algorithm, length, operation_policy_name=None, name=None):
             operation_policy_name (string): The name of the operation policy
                 to use for the new symmetric key. Optional, defaults to None
             name (string): The name to give the key. Optional, defaults to None
+            crypto_usage_mask (list): list of enumerations of crypto usage mask
+                passing to the symmetric key. Optional, defaults to empty list
```
PeterHamilton: Ditto.

kmip/pie/client.py
```diff
@@ -627,18 +631,21 @@ def mac(self, uid, algorithm, data):
...
         algorithm_attribute = self.attribute_factory.create_attribute(
             enums.AttributeType.CRYPTOGRAPHIC_ALGORITHM,
             algorithm)
         length_attribute = self.attribute_factory.create_attribute(
             enums.AttributeType.CRYPTOGRAPHIC_LENGTH,
             length)
+        # Default crypto usage mask value
+        mask_value = [enums.CryptographicUsageMask.ENCRYPT,
+                      enums.CryptographicUsageMask.DECRYPT]
+        mask_value.extend(masks)
```
PeterHamilton: We should probably remove any duplicates in the list here. It shouldn't really matter since the flags get combined latter on, but I'd prefer to keep things tidy as we go.

kmip/pie/client.py
```diff
@@ -138,7 +138,8 @@ def close(self):
                 self.logger.exception("could not close client connection", e)
                 raise e
 
-    def create(self, algorithm, length, operation_policy_name=None, name=None):
+    def create(self, algorithm, length, operation_policy_name=None, name=None,
+               cryptographic_usage_mask=list()):
```
PeterHamilton: Let's make the default value for `cryptographic_usage_mask` be `None`. There can be a nasty persistent state bug here if mutable objects are used as default values.

kmip/pie/client.py
```diff
@@ -149,6 +150,9 @@ def create(self, algorithm, length, operation_policy_name=None, name=None):
             operation_policy_name (string): The name of the operation policy
                 to use for the new symmetric key. Optional, defaults to None
             name (string): The name to give the key. Optional, defaults to None
+            cryptographic_usage_mask (list): list of enumerations of crypto
+                usage mask passing to the symmetric key. Optional, defaults
+                to empty list
```
PeterHamilton: This is fine, just change the default to `None`.

kmip/pie/client.py
```diff
@@ -627,18 +632,23 @@ def mac(self, uid, algorithm, data):
             message = result.result_message.value
             raise exceptions.KmipOperationFailure(status, reason, message)
 
-    def _build_key_attributes(self, algorithm, length):
+    def _build_key_attributes(self, algorithm, length, masks=list()):
```
PeterHamilton: Ditto.

kmip/tests/unit/pie/test_client.py
```diff
@@ -334,6 +383,19 @@ def test_create_on_invalid_length(self):
...
+    def test_create_on_invalid_cryptographic_usage_mask(self):
+        """
+        Test that a TypeError exception is raised when trying to create a
+        symmetric key with invalid cryptographic_usage_mask.
+        """
+        args = [enums.CryptographicAlgorithm.AES, 256]
+        kwargs = {'cryptographic_usage_mask':
+                  [enums.CryptographicUsageMask.ENCRYPT, 1]}
+        with ProxyKmipClient() as client:
+            self.assertRaises(TypeError, client.create, *args, **kwargs)
```
PeterHamilton: Let's also test if the `cryptographic_usage_mask` is not a list (e.g., someone just passes in a single enum value).

vbnmmnbv: Sure. But let me get clear. In this case, you want to raise an exception or not? Currently logic is that it will.

vbnmmnbv: Anyway. I have updated the codes~

PR #283, opened 2017-04-21
rename compromise_date to compromise_occurrence_date in revoke reques…

vbnmmnbv:
rename compromise_date to compromise_occurrence_date in revoke request payload

Both of them are valid attribute names and we should use the compromise_occurrence_date in revoke()

### DIFF
--- a/kmip/core/messages/payloads/revoke.py
+++ b/kmip/core/messages/payloads/revoke.py
@@ -40,7 +40,7 @@ class RevokeRequestPayload(Struct):
     def __init__(self,
                  unique_identifier=None,
                  revocation_reason=None,
-                 compromise_date=None):
+                 compromise_occurrence_date=None):
         """
         Construct a RevokeRequestPayload object.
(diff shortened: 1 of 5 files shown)


kmip/core/messages/payloads/revoke.py
```diff
@@ -40,21 +40,21 @@ class RevokeRequestPayload(Struct):
...
         Construct a RevokeRequestPayload object.
         Args:
             unique_identifier (UniqueIdentifier): The UUID of a managed
                 cryptographic object.
             revocation_reason (RevocationReason): The reason why the object was
                 revoked.
-            compromise_date (DateTime): the date of compromise if the object
-                was compromised.
+            compromise_occurrence_date (DateTime): the datetime when the object
+                was firstly believed to be compromised.
```
PeterHamilton: firstly -> first

vbnmmnbv:
Done~

PR #288, opened 2017-06-13
Adding request and response payloads for the Sign operation

danetrain:
This change adds the request and response payloads for the Sign operation.

### DIFF
--- a/kmip/tests/unit/core/messages/payloads/test_sign.py
+++ b/kmip/tests/unit/core/messages/payloads/test_sign.py
@@ -1,0 +1,782 @@
+# Copyright (c) 2017 The Johns Hopkins University/Applied Physics Laboratory
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import testtools
+
+from kmip.core import attributes
+from kmip.core import enums
+from kmip.core import utils
+
+from kmip.core.messages.payloads import sign
+
+
+class TestSignRequestPayload(testtools.TestCase):
+    """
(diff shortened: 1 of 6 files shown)


danetrain:
Peter, I'm getting some weird unit test failures locally. The read() method of the SignResponsePayload works fine if I call it once in a test, but when I call it twice in one unit test, the second call fails with the error:

struct.error: unpack requires a string argument of length 4

PeterHamilton:
It's been a while since I thought about this issue. The `BytearrayStream` object that we use for encodings is mutable. When the encoding gets read, it actually gets consumed by the reader. The bytes get pulled off the buffer and converted into whatever attribute or field is getting built at that moment, and the buffer shrinks. This makes it easy to track where we are in the buffer when reading; we're always at the beginning. When we use `is_tag_next`, we're peeking ahead into the buffer without pulling off bytes. When the reader is done reading/building itself, the buffer should be empty (that's what `is_oversized` checks for).

The simple fix is to just make a deep copy of the encoding and use that each time you want to read it in.

Long term, I've thought about changing how the `BytearrayStream` behaves (e.g., turning it into a static buffer). To be honest, it's not usually a problem. It's one of those library internal details that I just know and no one else thinks about.

danetrain:
Ok, I will update these tests to create a new Bytearraystream for each test.

kmip/tests/unit/core/messages/payloads/test_sign.py
```diff
@@ -0,0 +1,752 @@
...
+
+        payload = sign.SignResponsePayload()
+        args = (self.incomplete_encoding, )
+        self.assertRaisesRegexp(
+            ValueError,
+            "invalid payload missing the signature data attribute",
+            payload.read,
+            *args
+        )
+
```
PeterHamilton: The `read` tests look good but there need to be corresponding `write` tests as well.

danetrain:
Added write() tests for the response payload.

kmip/tests/unit/core/messages/payloads/test_sign.py
```diff
@@ -0,0 +1,782 @@
...
+    def test_not_equal_on_type_mismatch(self):
+        """
+        Test that the inequality operator returns True when comparing two
+        Sign request payloads with different types.
+        """
+        a = sign.SignRequestPayload()
+        b = 'invalid'
+
+        self.assertFalse(a == b)
+        self.assertFalse(b == a)
```
PeterHamilton: Update these two asserts to use `!=` instead.

PR #298, opened 2017-06-30
Add Name attribute filtering of locate for server

vbnmmnbv:
Originally I want to add the filtering logics for all attributes but found it's not trivial(is it?).
So for now I just add that for 'Name' attributes to avoid any surprise, while keeping in mind that we may want to add other attributes later.


PeterHamilton:
Nope, it's not trivial. That's why I've held off on it for so long.

Thanks for this, I'll try to give it a review this weekend.

PR #316, opened 2017-08-04
Add engine support for sign operation

danetrain:
This change adds the sign operation functionality
to the cryptography engine.

### DIFF
--- a/kmip/services/server/crypto/engine.py
+++ b/kmip/services/server/crypto/engine.py
@@ -13,6 +13,7 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+import binascii
 import logging
 import os
 
@@ -89,7 +90,8 @@ class CryptographyEngine(api.CryptographicEngine):
         }
         self._asymmetric_padding_methods = {
              enums.PaddingMethod.OAEP:     asymmetric_padding.OAEP,
-             enums.PaddingMethod.PKCS1v15: asymmetric_padding.PKCS1v15
+             enums.PaddingMethod.PKCS1v15: asymmetric_padding.PKCS1v15,
+             enums.PaddingMethod.PSS:      asymmetric_padding.PSS
         }
         self._symmetric_padding_methods = {
             enums.PaddingMethod.ANSI_X923: symmetric_padding.ANSIX923,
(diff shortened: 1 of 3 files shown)


kmip/services/server/crypto/engine.py
```diff
@@ -868,3 +884,123 @@ def wrap_key(self,
...
             )
+
+    def _create_RSA_private_key(self,
+                                bytes):
+        """
+        Instantiates an RSA key from bytes.
+
+        Args:
+            bytes (byte string): Bytes of public or private RSA key.
+            private (boolean): Indicates whether public or private key.
```
PeterHamilton: The `private` argument can be removed here, unless you plan on adding that capability to the method.

kmip/services/server/crypto/engine.py
```diff
@@ -868,3 +884,123 @@ def wrap_key(self,
...
+            (hash_alg, crypto_alg) = self._digital_signature_algorithms.get(
+                                         digital_signature_algorithm,
+                                         (None, None)
+            )
+
+        elif crypto_alg and hash_algorithm:
+            hash_alg = self._hash_algorithms.get(hash_algorithm, None)
+        else:
+            raise exceptions.InvalidField(
+                'For signing, either a digital signature  algorithm or a hash'
```
PeterHamilton: Nit: there's an extra space between `signature` and `algorithm`.

kmip/services/server/crypto/engine.py
```diff
@@ -868,3 +884,123 @@ def wrap_key(self,
...
+                    salt_length=asymmetric_padding.PSS.MAX_LENGTH
+                ),
+                hash_alg
+            )
+        else:
+            signature = key.sign(
+                data,
+                padding_method(hash_alg),
+                hash_alg
+            )
```
PeterHamilton: Given `OAEP` is also a supported asymmetric padding method, let's explicitly check that the `else` case matches `PKCS1v15` and then raise an error if it's anything else.

PeterHamilton:
Looks like this passes all the tests. Is this ready for a final review?

danetrain:
I believe so. I think I've done all the requested changes, and we've got 100% coverage w/ all unit tests passing.

PR #331, opened 2017-08-30
Add REVOKE to list of supported server ops

alishamayor:
As defined in _process_query, the list of supported server operations is missing REVOKE.


alishamayor:
I've gone ahead and committed the changes you requested.

PeterHamilton:
Thanks @alishamayor!

PR #364, opened 2017-11-13
add config about tpm

guo21:
Add some steps to config TPM.


PeterHamilton:
While this is a decent first stab at some config options for the TPM work, I think we should hold off on merging directly into ``master``. Go ahead and put together a working, end-to-end prototype in your own forked repo. That way you can experiment and tweak things without needing my approval for your design. Once you have a solid prototype, we can step back and take a look at what it would require to merge into ``master``.

PR #447, opened 2018-06-22
Clean up KMIPProxy initialization and destruction

### LINKED ISSUE #381: Bug: socket.error during test_kmip_client.py test execution
Running the unit test suite will sometimes generate the following error when executing `kmip/tests/unit/services/test_kmip_client.py`:
```
Exception socket.error: error(107, 'Transport endpoint is not connected') in <bound method
KMIPProxy.__del__ of <kmip.services.kmip_client.KMIPProxy object at 0x7f495986fed0>> ignored
```
Note that this is an intermittent error and shows up in roughly 1 in 20 test runs.

Identify the test causing this error and fix it.

tipabu:
Fixes #381.

### DIFF
--- a/kmip/services/kmip_client.py
+++ b/kmip/services/kmip_client.py
@@ -79,6 +79,10 @@ class KMIPProxy:
         self.logger = logging.getLogger(__name__)
         self.credential_factory = CredentialFactory()
         self.config = config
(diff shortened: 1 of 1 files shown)


PeterHamilton:
@tipabu So that's what was tripping the ``__del__`` error. Thanks for finding and fixing this!

PR #476, opened 2018-12-14
Update the PyKMIP object hierarchy to store the KMIP version

PeterHamilton:
This change updates the PyKMIP object hierarchy allowing it to store the KMIP version for which instantiated objects should be encoded and/or decoded. This is primarily a change in object constructors and base classes and does not greatly impact the existing object test infrastructure.


PeterHamilton:
Closing this in favor of #485

PR #497, opened 2019-02-26
Provision CA extension in CA cert

smira:
Without that, on modern OpenSSL libs this gives an error:

```
verify error:num=24:invalid CA certificate
```


PeterHamilton:
Hi @smira, thanks for submitting this pull request. It looks good to me!

PR #498, opened 2019-02-28
Move TLS server handshake down to connection thread

smira:
When TLS handshake is performed while in `accept()` call, main thread
might blocked up to network timeout effectively locking out other
clients from being able to establish connection with PyKMIP server.

Easy way to reproduce the problem:

1. Start PyKMIP server

2. Establish TCP connection with `nc -v 127.0.0.1 5696`

3. Attempt to connect (concurrently):
   `openssl s_client -host 127.0.0.1 -port 5696`

Without the fix, `openssl` would be blocked (won't even do initial TLS
handshake) until `nc` connection times out.

### DIFF
--- a/kmip/services/server/session.py
+++ b/kmip/services/server/session.py
@@ -97,14 +97,20 @@ class KmipSession(threading.Thread):
         """
         self._logger.info("Starting session: {0}".format(self.name))
 
-        while True:
-            try:
-                self._handle_message_loop()
-            except exceptions.ConnectionClosed as e:
-                break
-            except Exception as e:
-                self._logger.info("Failure handling message loop")
-                self._logger.exception(e)
+        try:
+            self._connection.do_handshake()
+        except Exception as e:
+            self._logger.info("Failure running TLS handshake")
+            self._logger.exception(e)
+        else:
+            while True:
+                try:
+                    self._handle_message_loop()
+                except exceptions.ConnectionClosed as e:
+                    break
+                except Exception as e:
+                    self._logger.info("Failure handling message loop")
+                    self._logger.exception(e)
 
         self._connection.shutdown(socket.SHUT_RDWR)
         self._connection.close()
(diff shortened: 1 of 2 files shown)


PeterHamilton:
Hi @smira, thanks for filing this issue. I've tested and reproduced the issue using the steps you provided. Thank you for those! I'm going to poke a bit more around the how the server handles the TLS socket but ultimately I think you're fix should work. I'll probably merge it in tomorrow.

kmip/services/server/session.py
```diff
@@ -97,6 +97,8 @@ def run(self):
         """
         self._logger.info("Starting session: {0}".format(self.name))
 
+        self._connection.do_handshake()
```
PeterHamilton: This should be wrapped in a try/except/else block in case the handshake fails. The except block should look something like L107-109 so that we log the handshake failure. The else block should just contain the while-loop at L102. If the handshake fails, we log the failure, skip the while-loop, and go straight to cleaning up the connection and closing out the session.

smira: not related to this PR, @PeterHamilton I also packaged some pretty trivial Docker image, are you interested in that? I can PR it

smira: thanks, updated

smira:
```
2019-03-01 16:19:22,277 - kmip.server - INFO - Receiving incoming connection from: 192.168.96.1:58906
2019-03-01 16:19:22,277 - kmip.server - INFO - Dedicating session 00000001 to 192.168.96.1:58906
2019-03-01 16:19:22,277 - kmip.server.session.00000001 - INFO - Starting session: 00000001
2019-03-01 16:19:32,287 - kmip.server.session.00000001 - INFO - Failure running TLS handshake
2019-03-01 16:19:32,287 - kmip.server.session.00000001 - ERROR - _ssl.c:1039: The handshake operation timed out
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/PyKMIP-0.9.dev0-py3.7.egg/kmip/services/server/session.py", line 101, in run
    self._connection.do_handshake()
  File "/usr/local/lib/python3.7/ssl.py", line 1117, in do_handshake
    self._sslobj.do_handshake()
socket.timeout: _ssl.c:1039: The handshake operation timed out
2019-03-01 16:19:32,288 - kmip.server.session.00000001 - INFO - Stopping session: 00000001
```

PeterHamilton:
Thanks for updating the patch. I've tested the changes and they work great. Merging now!

PR #572, opened 2019-08-07
Add support for AES GCM mode.

oleksiys:
KMIP 1.4 adds encryption/decryption parameters required by AES GCM mode.


PeterHamilton:
Hi @oleksiys, thanks for submitting this pull request! From my initial glance I'm impressed at how comprehensive it looks. I'll give it a deeper review within the next day or two.

PeterHamilton:
It looks like one of my tests is failing independent of your patch here. I'm triaging it now and should hopefully have a fix out tomorrow. Once that's upstream we can rebase your patch on that one and ensure that the test suite passes without issue.

PeterHamilton:
Now that #573 is merged, go ahead and rebase this and we'll verify the tests pass.

oleksiys:
PTAL

kmip/tests/unit/services/server/crypto/test_engine.py
```diff
@@ -1331,6 +1331,34 @@ def test_verify_signature_invalid_signing_algorithm(self):
...
+         ),
+         'auth_tag_length': 16,
+         'cipher_text': (
+             b'\x09\x45\x1b\x14\xf9\x25\x68\x24'
+             b'\x55\x50\x13\xb5\x09\xe9\xd6\x63'
+         ),
+         'auth_tag': (
+             b'\xbe\x9f\xef\x78\x5f\x9e\xe6\x16'
+             b'\x90\xe9\x44\x59\xc1\x84\x44\x7f'
+         )},
```
PeterHamilton: Where does this test vector come from?

kmip/services/server/crypto/engine.py
```diff
@@ -765,13 +834,18 @@ def _decrypt_symmetric(
                     raise exceptions.InvalidField(
                         "IV/nonce is required."
                     )
-                mode = mode(iv_nonce)
+                if is_gcm_mode:
+                    mode = mode(iv_nonce, auth_tag)
```
PeterHamilton: Let's use `tag=auth_tag` here, just to be explicit.

kmip/services/server/crypto/engine.py
```diff
@@ -459,7 +492,10 @@ def _encrypt_symmetric(
                 if iv_nonce is None:
                     iv_nonce = os.urandom(algorithm.block_size // 8)
                     return_iv_nonce = True
-                mode = mode(iv_nonce)
+                if is_gcm_mode:
+                    mode = mode(iv_nonce, None, auth_tag_length)
```
PeterHamilton: Let's use `tag=None` and `min_tag_length=auth_tag_length` here, just to be explicit.

kmip/tests/unit/services/server/crypto/test_engine.py
oleksiys: I just ran PyKMIP encode in GCM mode with pre-defined nonce, got this value as a result and recorded it.

PeterHamilton: That's what I thought. For this we need to test with known valid GCM test vectors, to prove that encrypt and decrypt work correctly with GCM. Many of the other test vectors in `test_engine.py` are pulled from existing NIST or IETF sources (see [here](https://github.com/OpenKMIP/PyKMIP/blob/b7a4e9be120185af1e3c819797416cd5c8b082c9/kmip/tests/unit/services/server/crypto/test_engine.py#L1786)). GCM test vectors should be available from NIST as well (see [here](https://csrc.nist.gov/Projects/cryptographic-algorithm-validation-program/CAVP-TESTING-BLOCK-CIPHER-MODES#GCMVS)), specifically the [Validation System document](https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Algorithm-Validation-Program/documents/mac/gcmvs.pdf) and the [test vector archive](http://csrc.nist.gov/groups/STM/cavp/documents/mac/gcmtestvectors.zip). It looks like [pyca/cryptography](https://github.com/pyca/cryptography) also has a [copy](https://github.com/pyca/cryptography/tree/master/vectors/cryptography_vectors/ciphers/AES/GCM) of the test vector archive, but I'd pull directly from the original source.

Take a look at how other tests handle the test vectors and pull in one or more test vectors from the NIST archive, replacing the test vector you have above. Be sure to add a comment pointing specifically to where you got the test vector. Once that's in place and the tests all pass, this should be good to merge.

oleksiys:
@PeterHamilton I migrated two tests from CAVP. I also realized that we need to strip the tag to the requested tag length, it's covered by the test.
PTAL.

PeterHamilton:
@oleksiys Thanks for updating this. I should be able to give this a final look over later this afternoon or tomorrow morning, at the latest.

kmip/tests/unit/services/server/crypto/test_engine.py
PeterHamilton: Thanks, this test case looks good to me.

kmip/tests/unit/services/server/crypto/test_engine.py
```diff
@@ -1387,6 +1387,34 @@ def test_verify_signature_invalid_signing_algorithm(self):
...
+         ),
+         'plain_text': (
+             b'\x8b\x4f\x7e\x75\x16\x31\xe7\x65'
+             b'\xdc\x13\xfa\x63\xf0\x2f\x63\x4b'
+         ),
+         'auth_additional_data': (
+             b'\x90\xee\x7e\x56\xf9\x59\x34\x76'
+             b'\x1c\x39\xab\x75\x37\x2a\xc2\xc6'
+         ),
+         'auth_tag_length': 8,
```
PeterHamilton: After looking through the `pyca/cryptography` source, `auth_tag_length` is definitely supposed to be bytes (so 8 here is correct). Please update the above GCM test vector (L1376) and change the `auth_tag_length` there from 128 (bits) down to 16 (bytes).

oleksiys: Good catch! Fixed.

Issue #450, opened 2018-07-30
Require "enum-compat" instead of "enum34"


staticglobal:
The enum34 package is only necessary for Python < 3.4, yet this package states it as a hard requirement, which is used no matter which version of Python is running.  This might not be a big deal except enum34 fails to install on newer versions of python when being built from source
```
$ pip install --user --no-binary :all: pykmip
Collecting pykmip
  Using cached https://files.pythonhosted.org/packages/48/2d/2a18894afb0666f6b273814e5e89460bfb63d0569ee35bcf4006505f39df/PyKMIP-0.8.0.tar.gz
Requirement already satisfied: cryptography in /usr/lib64/python3.6/site-packages (from pykmip)
Collecting enum34 (from pykmip)
  Using cached https://files.pythonhosted.org/packages/bf/3e/31d502c25302814a7c2f1d3959d2a3b3f78e509002ba91aea64993936876/enum34-1.1.6.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/usr/lib64/python3.6/site-packages/setuptools/__init__.py", line 5, in <module>
        import distutils.core
      File "/usr/lib64/python3.6/distutils/core.py", line 16, in <module>
        from distutils.dist import Distribution
      File "/usr/lib64/python3.6/distutils/dist.py", line 9, in <module>
        import re
      File "/usr/lib64/python3.6/re.py", line 142, in <module>
        class RegexFlag(enum.IntFlag):
    AttributeError: module 'enum' has no attribute 'IntFlag'
```

The "enum-compat" virtual package exists for this very reason.  Installs enum34 on older versions of Python, no-op for 3.4+.  Please change setup.py and requirements.txt to use this instead.

PeterHamilton:
@staticglobal Thanks for this. I wasn't aware of ``enum-compat``. I'll update requirements accordingly.

PR #579, opened 2019-08-28
Require enum-compat instead of enum34

ctil:
The enum34 package is not compatible with python 3.4+. By
requiring enum-compat instead, this requirement will be a
noop when installing the dependencies from a python 3.4+ environment.

Closes: #450

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,7 @@ setuptools.setup(
     },
     install_requires=[
         "cryptography",
-        "enum34",
+        "enum-compat",
         "requests",
         "six",
         "sqlalchemy"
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -50,7 +50,7 @@
     },
     install_requires=[
         "cryptography",
-        "enum34",
+        "enum-compat",
```
apevec, line 53: https://github.com/jstasiak/enum-compat/blob/master/README.rst suggests to use
enum34; python_version < '3.4'
directly.
With enum-compat, we get autogenerated RPM dep in Fedora Requires: python3.9dist(enum-compat) which does not exist.

ctil: Ah, looks like it should be changed back to `enum34; python_version < '3.4'`. Looks like the README in enum-compat was updated a couple months after this was merged to say it's no longer needed.

PR #635, opened 2020-09-15
Set the except clause as a parenthesized tuple

jcapiitao:
Hi folks,

In Fedora, the package build fails [1] for the current latest version (0.10.0), because python3.9 is used as default.
I know py39 is not yet supported, but this change makes the build process successful.

[1] https://koji.fedoraproject.org/koji/taskinfo?taskID=51511749

### DIFF
--- a/kmip/demos/pie/locate.py
+++ b/kmip/demos/pie/locate.py
@@ -63,7 +63,7 @@ if __name__ == '__main__':
     for initial_date in initial_dates:
         try:
             t = time.strptime(initial_date)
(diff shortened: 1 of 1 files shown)


PeterHamilton:
Hi @jcapiitao, thanks for submitting this patch. It's a minor change that I'm fine with merging in. Our CI system is currently broken due to older versions of Python breaking on different build hosts and incompatible dependencies. I need to go through and update our current set of supported environments.

Regardless, this looks fine to me. Thanks!

PR #650, opened 2021-07-22
Allow custom attributes

locula:
Add the ability to create and store custom attributes for SecretData objects and ensure that they are stored on the KMIP server when the register function is called

PR #654, opened 2021-09-08
Add custom attribute support for SymmetricKey object

locula:
I previously opened [Allow custom attributes #650](https://github.com/OpenKMIP/PyKMIP/pull/650) to add custom attribute support for SecretData objects. This PR is to add the same support for Symmetric Keys.

### DIFF
--- a/kmip/pie/objects.py
+++ b/kmip/pie/objects.py
@@ -658,7 +658,8 @@ class SymmetricKey(Key):
     }
 
     def __init__(self, algorithm, length, value, masks=None,
-                 name='Symmetric Key', key_wrapping_data=None, app_specific_info=None):
+                 name='Symmetric Key', key_wrapping_data=None,
+                 app_specific_info=None):
         """
         Create a SymmetricKey.
 
@@ -674,8 +675,9 @@ class SymmetricKey(Key):
             key_wrapping_data(dict): A dictionary containing key wrapping data
                 settings, describing how the key value has been wrapped.
                 Optional, defaults to None.
-            app_specific_info(list): A list of dictionaries containing application_namespace and application_data.
-                Optional, defaults to None.
+            app_specific_info(list): A list of dictionaries containing
+                application_namespace and application_data. Optional, defaults
+                to None.
         """
         super(SymmetricKey, self).__init__(
             key_wrapping_data=key_wrapping_data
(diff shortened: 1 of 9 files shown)


locula:
@PeterHamilton could you please take a look at this?

locula:
@PeterHamilton following up on this since this functionality is still being used, and it would be very helpful to get this merged. The changes are almost identical to those from https://github.com/OpenKMIP/PyKMIP/pull/650 which was merged.

kmip/pie/client.py
```diff
@@ -20,6 +20,7 @@
 from kmip.core import primitives
 from kmip.core import objects as cobjects
 
+from kmip.core.objects import Attribute
```
arp102, line 23: This looks like an unused import.

kmip/pie/objects.py
```diff
@@ -821,7 +824,7 @@ class PublicKey(Key):
 
     def __init__(self, algorithm, length, value,
                  format_type=enums.KeyFormatType.X_509, masks=None,
-                 name='Public Key', key_wrapping_data=None):
+                 name='Public Key', app_specific_info = None, key_wrapping_data=None):
```
arp102, line 827: There shouldn't be any whitespace around the `=`.

kmip/pie/objects.py
```diff
@@ -986,7 +992,7 @@ class PrivateKey(Key):
     }
 
     def __init__(self, algorithm, length, value, format_type, masks=None,
-                 name='Private Key', key_wrapping_data=None):
+                 name='Private Key', app_specific_info = None, key_wrapping_data=None):
```
arp102, line 995: There shouldn't be any whitespace around the =.

kmip/pie/objects.py
```diff
@@ -657,7 +657,7 @@ class SymmetricKey(Key):
         'sqlite_autoincrement': True
     }
 
-    def __init__(self, algorithm, length, value, masks=None,
+    def __init__(self, algorithm, length, value, app_specific_info=None, masks=None,
```
arp102, line 660: The position of the `app_specific_info` parameter is not consistent between the four init methods in #654 and #650.
It would be better to make it consistent, for example by making it the last parameter in all four methods.

arp102 (changes requested):
Hi, thanks for the pull request.
Sorry for not responding sooner.

The functionality looks good, but I've noted a few very minor quality issues.

Would you also be willing to add any unit tests that use the `app_specific_info`?

Thanks!

PR #656, opened 2021-09-22
Fix tests to pass with SQLAlchemy>=1.4.0

### LINKED ISSUE #649: Multiple tests failed one on particular filter construct.
When running the test suite for PyKMIP (using platform linux -- Python 3.6.13, pytest-6.2.4, py-1.10.0, pluggy-0.13.1, SQLAlchemy-1.4.15-207.2, [full log with all details](https://github.com/OpenKMIP/PyKMIP/files/6602386/_log.txt)) 46 tests fail on this construct:

```python
        get_obj = session.query(OpaqueObject).filter(
            ManagedObject.unique_identifier == obj.unique_identifier
            ).one()
```
with the error: ``sqlalchemy.orm.exc.DetachedInstanceError: Instance <OpaqueObject at 0x7f3da8148048> is not bound to a Session; attribute refresh operation cannot proceed (Background on this error at: http://sqlalche.me/e/14/bhk3)``. I was looking at the suggested web page, I dimly understand the issue, but I cannot understand how it applies here. Is it ``ManagedObject`` which is the unadjusted one, or ``obj``?

tipabu:
I'm not *entirely* sure what's going on here, but it seems that when we do something like

    obj = OpaqueObject(...)
    Session = sessionmaker(...)
    session = Session()
    ...
    session.add(obj)
    session.commit()

the primary key (and maybe some foreign relations?) aren't automatically populated on `obj` following the commit, and will attempt to lazy-load on next reference. Since `expire_on_commit` defaults to True, the session attached to `obj` (which is no longer the `session` in locals!) is closed out when we later do

    session = Session()
    get_obj = session.query(OpaqueObject).filter(
        ManagedObject.unique_identifier == obj.unique_identifier).one()

leading to a DetachedInstanceError.

There seem to be a few different ways we can fix this:

* Set `expire_on_commit=False` so the old session is still useful for the
  lazy-loading.
* Re-use the same session instead of creating a new one.
* Explicitly refresh added objects post-commit.

Generally prefer the first one; there's some prior art to follow in services/server/test_engine.py. Curiously, that same file runs into trouble despite already setting expire_on_commit=False -- so do the explicit refresh, on the assumption that there was a reason we went to the trouble of creating a fresh session.

Closes #649

### DIFF
--- a/kmip/tests/unit/pie/objects/test_opaque_object.py
+++ b/kmip/tests/unit/pie/objects/test_opaque_object.py
@@ -224,7 +224,7 @@ class TestOpaqueObject(testtools.TestCase):
         test_name = 'bowser'
         obj = OpaqueObject(
             self.bytes_a, enums.OpaqueDataType.NONE, name=test_name)
-        Session = sessionmaker(bind=self.engine)
+        Session = sessionmaker(bind=self.engine, expire_on_commit=False)
         session = Session()
         session.add(obj)
         session.commit()
@@ -259,7 +259,7 @@ class TestOpaqueObject(testtools.TestCase):
             expected_mo_names.append(sqltypes.ManagedObjectName(name, i))
         self.assertEquals(expected_mo_names, obj._names)
 
-        Session = sessionmaker(bind=self.engine)
+        Session = sessionmaker(bind=self.engine, expire_on_commit=False)
         session = Session()
         session.add(obj)
         session.commit()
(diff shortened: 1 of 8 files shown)


PeterHamilton (commented):
Overall I think adding `expire_on_commit=False` is a fine solution here. Adding `refresh` for those objects that need them is also fine.

PeterHamilton (approved):

PR #670, opened 2022-05-17
Repair some demos

### LINKED ISSUE #669: Some demos are broken
At present, most of the `register_*` demos crash with the `TypeError: build_cli_parser() missing 1 required positional argument: 'operation'` error. Additionally, `register_secret_data.py` incorrectly crates the `kmip.pie.objects.SecretData` object, which leads to the crash with the `TypeError: secret data mask (0 in list) must be a CryptographicUsageMask enumeration` error.

ktrushin:
Pass the required argument to the `build_cli_parser` function where
it was missed. Pass the missed argument in the `objects.SecretData`
initialization.
Fixes #669.


arp102:
Looks good, thanks!

PR #686, opened 2022-11-18
Encrypt kmip-server database using sqlcipher3.

janfrode:
Database will be encrypted by default, with password "pykmip". Password can be configured with the database_password in /etc/pykmip/server.conf.


janfrode:
It adds requirement for the python sqlcipher3 module plus some native packages (yum install sqlcipher-devel sqlcipher), or maybe just the sqlcipher3-binary module from PIP.

PR #689, opened 2023-01-26
Add cryptography >= 39.0.0 support

danigm:
The cryptography release 39.0.0 added a new parameter to the serializer that's required.

https://cryptography.io/en/latest/changelog/#v39-0-0

This patch fixes the tests test_encrypt_decrypt_asymmetric

### DIFF
--- a/kmip/services/server/crypto/engine.py
+++ b/kmip/services/server/crypto/engine.py
@@ -929,18 +929,18 @@ class CryptographyEngine(api.CryptographicEngine):
                     "decryption.".format(padding_method)
                 )
 
-            backend = default_backend()
-
             try:
-                private_key = backend.load_der_private_key(
+                private_key = serialization.load_der_private_key(
                     decryption_key,
-                    None
+                    password=None,
+                    backend=default_backend()
                 )
             except Exception:
                 try:
-                    private_key = backend.load_pem_private_key(
+                    private_key = serialization.load_pem_private_key(
                         decryption_key,
-                        None
+                        password=None,
+                        backend=default_backend()
                     )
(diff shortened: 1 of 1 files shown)


arp102:
I was confused why adding this argument was necessary when the cryptography documentation says that it's an optional argument that defaults to `False`.
(https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization/#cryptography.hazmat.primitives.serialization.load_pem_private_key)

The explanation seems to be that pykmip's `crypto/engine.py` is using two different cryptography functions named `load_pem_private_key`.
The function defined in `src/cryptography/hazmat/primitives/serialization/base.py` makes the new `unsafe_skip_rsa_key_validation` optional, but the function defined in `src/cryptography/hazmat/backends/openssl/rsa.py` makes it a required positional argument.
(See https://github.com/pyca/cryptography/commit/01687d63a5cb4475d84feae959bec6dea5fce3f1#diff-ede373fa44e8b647f096e9680b66759fb84ed5a5d112fd1fb65226b860aa1c56)

It seems to me that a better solution would be to change the `backend.load_pem_private_key` and `backend.load_der_private_key` calls (https://github.com/OpenKMIP/PyKMIP/blob/master/kmip/services/server/crypto/engine.py#L941) to match the `serialization.load_pem_private_key` and `serialization.load_der_private_key` calls (https://github.com/OpenKMIP/PyKMIP/blob/master/kmip/services/server/crypto/engine.py#L1283).
That seems to be more in line with how cryptography is intended to be used, and it should fix the exception without requiring us to check the cryptography version number or do anything with the new argument.

PR #703, opened 2023-08-25
Add automatic tests with Github Actions

justin-h-loi:
Current setup:
- Runs on push/pull_request to master
- Images: Ubuntu-20.04:py38 and Ubuntu-22.04:py10
- Tests: tox unit and integration tests


.github/workflows/.travis.yml
```diff
@@ -0,0 +1,42 @@
+name: ".travis.yml"
```
arp102, line 1: Having "travis" in the file name doesn't make sense because we aren't using travis any more, and I don't think the file should be hidden either, so I would rename it to `tox.yml` (both this name field and the file itself).

.github/workflows/.travis.yml
```diff
@@ -0,0 +1,42 @@
...
+            RUN_INTEGRATION_TESTS: ${{ matrix.test_number}}
+        steps:
+            - uses: actions/checkout@v3
+            - run: pip install "tox"
+            - run: pip install bandit
+            - run: pip install codecov
+            - run: pip install slugs
+            - run: sudo chmod -R 777 /usr/local/lib/
+            - run: python3 setup.py install
+            - run: RUN_INTEGRATION_TESTS=${{ matrix.test_number }} ./.travis/run.sh
```
arp102, line 23: The variable here seems redundant with the variable definition above.

```suggestion
            - run: ./.travis/run.sh
```

.github/workflows/.travis.yml
```diff
@@ -0,0 +1,42 @@
...
+            - run: codecov
+        strategy:
+            matrix:
+                test: ["pep8", "bandit", "docs"]
+                test_number: ["0","1"]
```
arp102, line 28: The test_number doesn't affect the pep8, bandit, or docs tests, so we're running each of those tests twice for no reason.
It would be preferable to just remove this line, but I'm not sure if that would break the env section above.
If the tests break when `test_number` is undefined, then we should change this list to just have one entry.

kmip/services/server/session.py
```diff
@@ -127,7 +127,8 @@ def _handle_message_loop(self):
         )
 
         try:
-            if hasattr(self._connection, 'shared_ciphers') and self._connection.shared_ciphers() is not None:
+            if hasattr(self._connection, 'shared_ciphers') \
+                    and self._connection.shared_ciphers() is not None:
```
arp102, line 131: Parentheses are usually preferable to escaped newlines.

```suggestion
            if (hasattr(self._connection, 'shared_ciphers')
                    and self._connection.shared_ciphers() is not None):
```

kmip/services/server/crypto/engine.py
```diff
@@ -269,7 +269,8 @@ def mac(self, algorithm, key, data):
             )
             cipher_algorithm = self._symmetric_key_algorithms.get(algorithm)
             try:
-                # ARC4 and other non-block cipher algorithms will raise TypeError exceptions
+                # ARC4 and other non-block cipher algorithms will
+                # raise TypeError exceptions
```
arp102, line 273:
```suggestion
                # ARC4 and other non-block cipher algorithms will raise TypeError
```

PR #707, opened 2023-10-04
Add python 3.12 and 3.13 support

### LINKED ISSUE #706: Use of removed ssl.wrap_socket
`ssl.wrap_socket` was [removed in Python 3.12](https://docs.python.org/3.12/whatsnew/3.12.html#ssl) (it was deprecated in Python 3.7).  I'll make a PR to update.

blink1073:
Fixes #706

### DIFF
--- a/kmip/services/kmip_client.py
+++ b/kmip/services/kmip_client.py
@@ -285,11 +285,12 @@ class KMIPProxy(object):
             six.reraise(*last_error)
 
     def _create_socket(self, sock):
-        context = ssl.SSLContext(self.ssl_version)
+        context = ssl.SSLContext(self.auth_suite.protocol)
         context.verify_mode = self.cert_reqs
         if self.ca_certs:
             context.load_verify_locations(self.ca_certs)
-        context.check_hostname = False
+        if self.keyfile and not self.certfile:
+            raise ValueError("certfile must be specified")
         if self.certfile:
             context.load_cert_chain(self.certfile, self.keyfile)
         self.socket = context.wrap_socket(
--- a/kmip/services/kmip_client.py
+++ b/kmip/services/kmip_client.py
@@ -285,7 +285,7 @@ class KMIPProxy(object):
             six.reraise(*last_error)
 
     def _create_socket(self, sock):
-        context = ssl.SSLContext(self.auth_suite.protocol)
+        context = ssl.SSLContext(self.ssl_version)
         context.verify_mode = self.cert_reqs
         if self.ca_certs:
             context.load_verify_locations(self.ca_certs)
--- a/kmip/services/kmip_client.py
+++ b/kmip/services/kmip_client.py
@@ -285,7 +285,7 @@ class KMIPProxy(object):
             six.reraise(*last_error)
 
     def _create_socket(self, sock):
-        context = ssl.create_default_context(purpose=ssl.Purpose.SERVER_AUTH, cafile=self.certfile)
+        context = ssl.create_default_context(purpose=ssl.Purpose.SERVER_AUTH, capath=self.ca_certs)
         context.check_hostname = False
         context.load_cert_chain(self.certfile, self.keyfile)
         self.socket = context.wrap_socket(
--- a/kmip/services/kmip_client.py
+++ b/kmip/services/kmip_client.py
@@ -287,6 +287,7 @@ class KMIPProxy(object):
     def _create_socket(self, sock):
         context = ssl.create_default_context(purpose=ssl.Purpose.SERVER_AUTH, cafile=self.certfile)
         context.check_hostname = False
+        context.load_cert_chain(self.certfile, self.keyfile)
         self.socket = context.wrap_socket(
             sock,
             server_side=False,
--- a/kmip/services/kmip_client.py
+++ b/kmip/services/kmip_client.py
@@ -285,10 +285,12 @@ class KMIPProxy(object):
             six.reraise(*last_error)
 
     def _create_socket(self, sock):
-        context = ssl.create_default_context(purpose=ssl.Purpose.SERVER_AUTH, capath=self.ca_certs)
+        purpose = ssl.Purpose.SERVER_AUTH
+        context = ssl.create_default_context(purpose=purpose, capath=self.ca_certs)
         context.check_hostname = False
         context.load_default_certs(ssl.Purpose.SERVER_AUTH)
-        context.load_cert_chain(self.certfile, self.keyfile)
+        if self.certfile:
+            context.load_cert_chain(self.certfile, self.keyfile)
         self.socket = context.wrap_socket(
             sock,
             server_side=False,
--- a/kmip/services/kmip_client.py
+++ b/kmip/services/kmip_client.py
@@ -285,8 +285,10 @@ class KMIPProxy(object):
             six.reraise(*last_error)
 
     def _create_socket(self, sock):
-        purpose = ssl.Purpose.SERVER_AUTH
-        context = ssl.create_default_context(purpose=purpose, capath=self.ca_certs)
+        context = ssl.SSLContext(self.ssl_version)
+        context.verify_mode = self.cert_reqs
+        if self.ca_certs:
+            context.load_verify_locations(self.ca_certs)
         context.check_hostname = False
         if self.certfile:
             context.load_cert_chain(self.certfile, self.keyfile)
--- a/kmip/services/kmip_client.py
+++ b/kmip/services/kmip_client.py
@@ -288,7 +288,6 @@ class KMIPProxy(object):
         purpose = ssl.Purpose.SERVER_AUTH
         context = ssl.create_default_context(purpose=purpose, capath=self.ca_certs)
         context.check_hostname = False
-        context.load_default_certs(ssl.Purpose.SERVER_AUTH)
         if self.certfile:
             context.load_cert_chain(self.certfile, self.keyfile)
         self.socket = context.wrap_socket(
--- a/kmip/services/kmip_client.py
+++ b/kmip/services/kmip_client.py
@@ -285,16 +285,16 @@ class KMIPProxy(object):
             six.reraise(*last_error)
 
     def _create_socket(self, sock):
-        with ssl.SSLContext() as ctx:
-            self.socket = ctx.wrap_socket(
-                sock,
-                keyfile=self.keyfile,
-                certfile=self.certfile,
-                cert_reqs=self.cert_reqs,
-                ssl_version=self.ssl_version,
-                ca_certs=self.ca_certs,
-                do_handshake_on_connect=self.do_handshake_on_connect,
-                suppress_ragged_eofs=self.suppress_ragged_eofs)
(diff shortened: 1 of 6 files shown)


s-t-e-v-e-n-k:
The problem with your mocking is that you can't mock `ssl.SSLContext.wrap_socket` because the code constructs an `SSLContext` object and **then** calls `wrap_socket` on that, so you're not mocking the right thing.

I've edited the test changes for your patch (and not your code changes, which looks good) thusly:

```
@@ -210,9 +210,9 @@ class TestKmipServer(testtools.TestCase)
         # Test that in ideal cases no errors are generated and the right
         # log messages are.
         with mock.patch('socket.socket') as socket_mock:
-            with mock.patch('ssl.wrap_socket') as ssl_mock:
+            with mock.patch('ssl.SSLContext') as ssl_mock:
                 socket_mock.return_value = a_mock
-                ssl_mock.return_value = b_mock
+                ssl_mock.return_value.wrap_socket.return_value = b_mock

                 manager_mock.assert_not_called()
                 monitor_mock.assert_not_called()
```

And I've confirmed the tests pass with Python 3.9-3.12.

blink1073:
Thanks @s-t-e-v-e-n-k, new build in progress on my [fork](https://github.com/blink1073/PyKMIP/pull/1).

blink1073:
Still no joy: `ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate (_ssl.c:1133)`

avikivity:
@blink1073 were you able to make this work?

blink1073:
No, I haven't been able to figure out the testing portion.

kmip/services/kmip_client.py
petere, line 1: This no longer pays attention to the configuration settings `self.cert_reqs` and `self.ssl_versions`. That's probably okay in practice, since the defaults are good, but these settings are documented, so they should work.

One option might be to not use `ssl.create_default_context()` and instead use `ssl.SSLContext` directly and apply these settings. Like

```python
        context = ssl.SSLContext(self.ssl_version)
        context.verify_mode = self.cert_reqs
        if self.ca_certs:
            context.load_verify_locations(self.ca_certs)
```

This is just about as much code as before but it reproduces the previous behavior more precisely.

petere (changes requested):

kmip/services/kmip_client.py
```diff
@@ -285,13 +285,14 @@ def open(self):
             six.reraise(*last_error)
 
     def _create_socket(self, sock):
-        self.socket = ssl.wrap_socket(
+        purpose = ssl.Purpose.SERVER_AUTH
+        context = ssl.create_default_context(purpose=purpose, capath=self.ca_certs)
```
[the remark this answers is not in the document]
blink1073: Fixed

kmip/services/server/server.py
```diff
@@ -287,17 +287,20 @@ def interrupt_handler(trigger, frame):
         for cipher in auth_suite_ciphers:
             self._logger.debug(cipher)
 
-        self._socket = ssl.wrap_socket(
+        purpose = ssl.Purpose.CLIENT_AUTH
+        capath = self.config.settings.get('ca_path')
+        context = ssl.create_default_context(purpose=purpose, capath=capath)
```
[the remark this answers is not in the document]
blink1073: Fixed

kmip/services/kmip_client.py
blink1073: Done

kmip/services/server/server.py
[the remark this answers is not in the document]
blink1073: Updated

kmip/services/server/server.py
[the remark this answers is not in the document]
blink1073: Done

blink1073:
Thanks @petere!  I believe I've addressed your feedback.

kmip/services/kmip_client.py
```diff
@@ -285,13 +285,16 @@ def open(self):
             six.reraise(*last_error)
 
     def _create_socket(self, sock):
-        self.socket = ssl.wrap_socket(
+        context = ssl.SSLContext(self.ssl_version)
+        context.verify_mode = self.cert_reqs
+        if self.ca_certs:
+            context.load_verify_locations(self.ca_certs)
+        context.check_hostname = False
+        if self.certfile:
```
petere, line 293: I'm testing combinations of "forgetting" to specify `certfile` or `keyfile` in the configuration file. With this code, if you specify `certfile` but not `keyfile`, you'll get an exception from `load_cert_chain`. If you specify `keyfile` but not `certfile`, this code is skipped and you'll get whatever the SSL stack ends up deciding. The old module-level `wrap_socket` had an additional check

```python
    if keyfile and not certfile:
        raise ValueError("certfile must be specified")
```

to prevent this. Maybe this would be good to have here as well.

kmip/services/server/server.py
```diff
@@ -287,17 +287,24 @@ def interrupt_handler(trigger, frame):
         for cipher in auth_suite_ciphers:
             self._logger.debug(cipher)
 
-        self._socket = ssl.wrap_socket(
+        cafile = self.config.settings.get('ca_path')
+        context = ssl.SSLContext(self.ssl_version)
```
petere, line 291: This should be

```python
context = ssl.SSLContext(self.auth_suite.protocol)
```

(wrongly copy-pasted from client code?)

kmip/services/server/server.py
```diff
@@ -287,17 +287,24 @@ def interrupt_handler(trigger, frame):
...
+        context.verify_mode = ssl.CERT_REQUIRED
+        if auth_suite_ciphers:
```
petere, line 293: This should be `self.auth_suite.ciphers`. I'm not sure exactly what the relationship between that and `auth_suite_ciphers` is, but this one crashes.

kmip/services/server/server.py
```diff
@@ -287,17 +287,24 @@ def interrupt_handler(trigger, frame):
...
+            context.set_ciphers(auth_suite_ciphers)
+        if cafile:
+            context.load_verify_locations(cafile)
+        certfile = self.config.settings.get('certificate_path')
+
+        if certfile:
```
petere, line 299: As with the client, it's worth checking here about how different combinations of not specifying certificate or key behave. For example, the old module-level `wrap_socket()` had a check

```python
    if server_side and not certfile:
        raise ValueError("certfile must be specified for server-side "
                         "operations")
```

so omitting the certificate would have been a hard error.

petere (changes requested):

kmip/services/kmip_client.py
```diff
@@ -285,13 +285,16 @@ def open(self):
             six.reraise(*last_error)
 
     def _create_socket(self, sock):
-        self.socket = ssl.wrap_socket(
+        context = ssl.SSLContext(self.ssl_version)
+        context.verify_mode = self.cert_reqs
+        if self.ca_certs:
+            context.load_verify_locations(self.ca_certs)
+        context.check_hostname = False
```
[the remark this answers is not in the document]
blink1073: Removed

kmip/services/kmip_client.py
blink1073: Done

kmip/services/server/server.py
blink1073: Fixed

kmip/services/server/server.py
blink1073: Fixed

kmip/services/server/server.py
```diff
@@ -287,17 +287,26 @@ def interrupt_handler(trigger, frame):
...
+            context.set_ciphers(self.auth_suite.ciphers)
+        if cafile:
+            context.load_verify_locations(cafile)
+        certfile = self.config.settings.get('certificate_path')
+
+        if certfile:
+            keyfile = self.config.settings.get('key_path')
+            context.load_cert_chain(certfile, keyfile=keyfile)
+        else:
+            raise ValueError("certfile must be specified for server-side operations")
```
petere: What we don't have here is the equivalent of the client-side check

```python
        if self.keyfile and not self.certfile:
            raise ValueError("certfile must be specified")
```

but it also turns out that the higher levels already check that both certificate and key are specified:

```
kmip.core.exceptions.ConfigurationError: Setting 'key_path' is missing from the configuration file.
kmip.core.exceptions.ConfigurationError: Setting 'certificate_path' is missing from the configuration file.
```

So it's probably not necessary.

petere (approved):

kmip/services/server/server.py
petere: Also, it looks like this causes the test suite to fail in `TestKmipServer.test_start`:

```
ValueError: certfile must be specified for server-side operations
```

So either this code or the test setup would need adjustments.

blink1073: Removed from the server

blink1073:
@petere is there anything else you'd like me to try?

kmip/services/server/server.py
petere: I think what I had in mind was like

```diff
         if certfile:
             keyfile = self.config.settings.get('key_path')
             context.load_cert_chain(certfile, keyfile=keyfile)
+        else:
+            raise ValueError("certfile must be specified for server-side operations")
```

petere:
Other than the single comment I just added (which I just got my looking at my local copy, but it's been a while since I played with this), I am content with this version.

How we get from that to getting this merged and released, I don't know.

kmip/services/server/server.py
blink1073: Fixed

blink1073: I instead let the context raise an appropriate error.  This has the benefit of letting us mock the context behavior in the tests.

blink1073:
This is now ready for review.   cc @arp102.

blink1073:
The workflows are all passing on my fork: https://github.com/blink1073/PyKMIP/actions/runs/12119149288?pr=1

tomamoto:
@arp102 Are you able to merge this PR, and if not, what else is needed? Getting 3.12 support in master branch would be great, and seeing a new release to pypi would be even nicer.

BourgoisMickael (approved):

wdelgenio:
@arp102 @BourgoisMickael @petere @PeterHamilton 
Do any one of you have permissions to merge this fix?  This one is needed for anyone who is running python3.12 or newer and is a complete showstopper.

BourgoisMickael:
> @arp102 @BourgoisMickael @petere @PeterHamilton Do any one of you have permissions to merge this fix? This one is needed for anyone who is running python3.12 or newer and is a complete showstopper.

I'm not related to the project I just want to see it merged

wdelgenio:
Looks like this repo is no longer maintained, might be time for a fork or a switch to a different library.

dmendiza:
@wdelgenio Hi, I've volunteered to continue to maintain this repo.  I'm merging this PR and will make a release soon.

blink1073:
Great, thank you!

PR #715, opened 2024-03-29
Update supported python versions

kajinamit:
... according to the versions actually tested now.


rajivmucheli:
Hi @arp102, could you please review ? this will continue supporting barbican-kmip integration.

PR #720, opened 2024-07-27
Initial TLS 1.3 support

oittaa:


oittaa:
With all the mocking going on, I don't really know how to make the tests work at the moment. However I built a Docker image with Alpine Linux 3.20, which has Python 3.12, and it seemed to work.

PR #733, opened 2025-08-28
Fixed compatability with Python 3.12

ymeiron:
Python 3.12 removed the `ssl.wrap_socket()` function that was deprecated in Python 3.7, breaking both server and client functionality of PyKMIP. To fix the issue, we create an `ssl.SSLContext` object using `ssl.create_default_context()` and use its `wrap_socket()` instead.


ymeiron:
I see #707 already exists and never been merged

PR #739, opened 2026-08-31
trim more "six" usage

a-detiste:


a-detiste:
enum-compat should be dropped too (< Python 3.4 stuff)

